PR #35, opened 2015-02-09
Add 3.5.0a1

tianon:
This is literally just a copy of the 3.4 directory with all the version numbers changed.  Building it now to make sure it actually works. :+1:

### DIFF
--- a/3.5/Dockerfile
+++ b/3.5/Dockerfile
@@ -1,0 +1,43 @@
+FROM buildpack-deps:jessie
+
+# remove several traces of debian python
+RUN apt-get purge -y python.*
+
+# http://bugs.python.org/issue19846
+# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
+ENV LANG C.UTF-8
+
+ENV PYTHON_VERSION 3.5.0a1
+ENV PYTHON_VERSION_DIR 3.5.0
+
+# gpg: key F73C700D: public key "Larry Hastings <larry@hastings.org>" imported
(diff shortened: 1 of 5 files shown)


tianon:
This one needs to wait for #36, since it has to change the same bits (and propagate them in more places).  I'll rebase and update this after that one's reviewed and merged.

tianon:
Rebased! :tada:

yosifkit:
I'm not sure I like the extra `ENV` for the directory in the url.  It makes 3.5 different from the other versions of python.

tianon:
Technically, 3.4 should have the same thing (as I discovered with `3.4.3rc1` that we discussed briefly).  Would you feel better about it if I copied the same syntax to 3.4 and 3.3?  Or would you rather I embed it somehow in the actual build line?

yosifkit:
For 3.4 (or any version), we don't really want to push an alpha or RC unless there is not a real release for that major version, since we do not have an easy way to do both the release 3.4.2 and the pre-release 3.4.3rc1.

As far as how to reconcile this for all versions, we could move your logic from `update.sh` to the beginning of each download block `versionDir="$(echo "$PYTHON_VERSION" | sed -r 's/[^0-9.]+.*$//')"`, but if you think ENVs in all the versions is a better solution we can do that.  Really, I just want to ensure consistency where possible.

tianon:
This is a general problem we have with multiple repos we maintain, not just this one, and I don't feel like we have a good solution yet. :disappointed:

I agree that this isn't great, especially since it's inconsistent, so I'm closing for now.

PR #34, opened 2015-02-04
Gpg verification

Moghedrin:
Add gpg verification to the Python image.

Fixes #32

PR #36, opened 2015-02-09
Add GPG verification to slim and wheezy variants too

tianon:
Also, fix `update.sh` to actually hit both the slim and wheezy variants when bumping...

This is the bits that were missed in #34.


yosifkit:
LGTM

Issue #21, opened 2014-11-21
Python should be configured/built with --enable-shared option.


GrahamDumpleton:
Currently the Python installations are built with a command like:

```
RUN mkdir -p /usr/src/python \
    && curl -SL "https://www.python.org/ftp/python/$PYTHON_VERSION/Python-$PYTHON_VERSION.tar.xz" \
        | tar -xJC /usr/src/python --strip-components=1 \
    && cd /usr/src/python \
    && ./configure \
    && make -j$(nproc) \
    && make install \
    && cd / \
    && rm -rf /usr/src/python
```

When running `configure`, you should be supplying the `--enable-shared` option to ensure that shared libraries are built for Python. By not doing this you are preventing any application which wants to use Python as an embedded environment from working. This is because the lack of the shared library results in any embedded system failing at compile time with:

```
/usr/bin/ld: /usr/local/lib/libpython2.7.a(abstract.o): relocation R_X86_64_32S against `_Py_NotImplementedStruct' can not be used when making a shared object; recompile with -fPIC

/usr/local/lib/libpython2.7.a: error adding symbols: Bad value

collect2: error: ld returned 1 exit status

error: command 'gcc' failed with exit status 1
```

This basic mistake is something that Linux distributions themselves made for many years and it took a lot of complaining and education to get them to fix their Python installations. It would be nice to see you address this and do what all decent Linux distributions do now, and have done for a while, and install Python with shared libraries.

Starefossen:
> Next is the presence of --enable-unicode=ucs4. By default Python when built from source code uses ucs2 for Unicode. I don't really understand the Linux rational for using ucs4 but every Linux distribution I have seen overrides it and uses ucs4.

I ran into this issue that this version of Python isn't using `ucs4`. Is there any possiblility to append/override configuration options using environment variables like [pyenv](https://github.com/yyuu/pyenv/blob/master/plugins/python-build/README.md#special-environment-variables) does?

gjcarneiro:
Unfortunately --enable-shared builds a CPython interpreter that is [slower](https://gist.github.com/methane/4681131).  What distributions usually do is build Python twice, once static, to make the /usr/bin/python interpreter faster, and another time with --enable-shared just so that it installs the shared library.

GrahamDumpleton:
[...]

PR #38, opened 2015-02-26
Configure Python with --enable-unicode=ucs4

Starefossen:
Ref. #21 I went ahead and added the `--enable-unicode=ucs4` configuration flag.

Maybe this could be solved via a `PYTHON_CONFIGURE_OPTS` environment variable?

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -20,7 +20,7 @@ RUN set -x \
 	&& tar -xJC /usr/src/python --strip-components=1 -f python.tar.xz \
 	&& rm python.tar.xz* \
 	&& cd /usr/src/python \
-	&& ./configure --enable-shared \
+	&& ./configure --enable-shared --enable-unicode=ucs4 \
 	&& make -j$(nproc) \
 	&& make install \
 	&& ldconfig \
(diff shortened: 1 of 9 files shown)


tianon:
This seems reasonable to me. :+1:

@paultag since you have really strong opinions on Python and UTF8, care to weigh in?

https://docs.python.org/2/c-api/unicode.html
> This type represents the storage type which is used by Python internally as basis for holding Unicode ordinals. Python’s default builds use a 16-bit type for Py_UNICODE and store Unicode values internally as UCS2. It is also possible to build a UCS4 version of Python (most recent Linux distributions come with UCS4 builds of Python). These builds then use a 32-bit type for Py_UNICODE and store Unicode data internally as UCS4. On platforms where wchar_t is available and compatible with the chosen Python Unicode build variant, Py_UNICODE is a typedef alias for wchar_t to enhance native platform compatibility. On all other platforms, Py_UNICODE is a typedef alias for either unsigned short (UCS2) or unsigned long (UCS4).

https://docs.python.org/3/c-api/unicode.html also has this interesting tidbit: (but not much else one way or another)
> When dealing with single Unicode characters, use Py_UCS4.

paultag:
Euch, sorry, was going to reply, yeah this seems entirely reasonable to me!

:+1:

tianon:
LGTM

tianon:
(thanks @paultag :heart:)

yosifkit:
LGTM

Starefossen:
Does anyone know when the image on Docker Hub will be updated?

tianon:
See https://github.com/docker-library/official-images/pull/528 for progress
on that.

PR #41, opened 2015-04-02
 Add libbz2-dev to buildDeps

### LINKED ISSUE #39: Should slim also install libbz2-dev
When trying to `pip install pytz` i get the following error message:

    tarfile.CompressionError: bz2 module is not available

Imo libbz2-dev should be installed and included before building python.

denibertovic:
Closes #39

@tianon @yosifkit  I'm not sure what to do about the tests ... that's mentioned in the original issue. Is this good to go as is? (and then refactor the tests as needed?)

### DIFF
--- a/3.4/slim/Dockerfile
+++ b/3.4/slim/Dockerfile
@@ -19,7 +19,7 @@ ENV PYTHON_VERSION 3.4.3
 RUN gpg --keyserver pool.sks-keyservers.net --recv-keys 97FC712E4C024BBEA48A61ED3A5CA953F73C700D
 
 RUN set -x \
-	&& buildDeps='curl gcc libc6-dev libsqlite3-dev libssl-dev make xz-utils zlib1g-dev' \
+	&& buildDeps='curl gcc libbz2-dev libc6-dev libsqlite3-dev libssl-dev make xz-utils zlib1g-dev' \
 	&& apt-get update && apt-get install -y $buildDeps --no-install-recommends && rm -rf /var/lib/apt/lists/* \
 	&& mkdir -p /usr/src/python \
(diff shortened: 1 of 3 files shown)


3.4/slim/Dockerfile
```diff
@@ -19,7 +19,7 @@ ENV PYTHON_VERSION 3.4.3
 RUN gpg --keyserver pool.sks-keyservers.net --recv-keys 97FC712E4C024BBEA48A61ED3A5CA953F73C700D
 
 RUN set -x \
-	&& buildDeps='curl gcc libc6-dev libsqlite3-dev libssl-dev make xz-utils zlib1g-dev' \
+	&& buildDeps='curl gcc libc6-dev libsqlite3-dev libssl-dev make xz-utils zlib1g-devi libbz2-dev' \
```
tianon: `devi`? :+1:

(you must be a Vim user! :smile:)

Also, worth noting that these are sorted alphabetically. :wink:

tianon:
You've tested to make sure this actually _fixes_ #39, right, and aren't just taking our word that it's probably the right fix? :smile:

tianon:
(and yeah, adding a test to verify this in the future too to the official images tests would be a good idea, but orthogonal)

denibertovic:
@tianon yeah vim user. Sorry about that typo :) I'll address your comments later today.

Oh, and yes I've verified it actually fixes the issue because I've bumped into it myself, and after adding this it worked.

denibertovic:
I should mention that I used the 2.7 version.

tianon:
Cool, sounds good :+1:  Just give a shout when you've got the fixes in. :smile:

denibertovic:
@tianon @yosifkit updated! I will add tests in a separate PR for https://github.com/docker-library/official-images/issues/560

yosifkit:
LGTM

tianon:
LGTM

PR #44, opened 2015-04-22
More resilient gpg getting

yosifkit:
- move to "high-availability" [subset](https://sks-keyservers.net/overview-of-pools.php#pool_ha).

related: https://github.com/docker-library/php/pull/92


tianon:
LGTM

PR #46, opened 2015-05-19
Install explicit version of "pip"

tianon:


2.7/Dockerfile
```diff
@@ -12,6 +12,9 @@ RUN gpg --keyserver ha.pool.sks-keyservers.net --recv-keys C01E1CAD5EA2C4F0B8E35
 
 ENV PYTHON_VERSION 2.7.9
 
+# if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
+ENV _PIP_VERSION 6.1.1
```
Starefossen: I would maybe have preferred naming it something less awkward like `DOCKER_PIP_VERSION` or `PYTHON_PIP_VERSION`.

tianon: That's definitely easy to change. :+1:

Starefossen: Otherwise this PR looks solid to me :+1:

tianon: Updated! :heart:

yosifkit:
LGTM

PR #48, opened 2015-05-24
Add support for Python 3.2

radeksimko:


tianon:
Hmm, it would appear that 3.2 is actually still "supported" (at least for security updates) until February 2016 (https://www.python.org/dev/peps/pep-0392/#lifespan).

LGTM :+1:

yosifkit:
LGTM

radeksimko:
Thanks, since there doesn't seem to be any direct integration w/ Docker Hub, what's the process behind getting it to https://registry.hub.docker.com/_/python/ as a tag + mentioning it in the Readme (_Information_ tab)?

yosifkit:
It is a PR to the official-images, like this one: [docker-library/official-images#758](https://github.com/docker-library/official-images/pull/758/files).  The `library/python` file there is generated by running the `generate-stackbrew-library.sh` from this repo.

PR #50, opened 2015-06-05
Skip pip's default caching behavior

mattrobenolt:
This caching includes the file that was downloaded over http,
as well as this new wheel cache. Neither of which are valuable
and just add bloat into the resulting images. Similar to removing
the apt caches on disk at each layer. This also makes pip install's
a tiny bit faster since pip avoids trying to automatically create
wheels (which is fast itself, but again, unnecessary in this world).

See:
  - https://pip.pypa.io/en/latest/reference/pip_install.html#caching
  - https://pip.pypa.io/en/latest/reference/pip_install.html#wheel-cache

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -28,7 +28,7 @@ RUN set -x \
 	&& make install \
 	&& ldconfig \
 	&& curl -SL 'https://bootstrap.pypa.io/get-pip.py' | python2 \
-	&& pip install --upgrade pip==$PYTHON_PIP_VERSION \
+	&& pip install --no-cache-dir --upgrade pip==$PYTHON_PIP_VERSION \
 	&& find /usr/local \
 		\( -type d -a -name test -o -name tests \) \
 		-o \( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
@@ -36,6 +36,6 @@ RUN set -x \
 	&& rm -rf /usr/src/python
 
 # install "virtualenv", since the vast majority of users of this image will want it
-RUN pip install virtualenv
+RUN pip install --no-cache-dir virtualenv
 
 CMD ["python2"]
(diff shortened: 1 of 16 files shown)


yosifkit:
Great idea! How about just the following so that it applies to all pip invocations?

```Dockerfile
RUN { echo '[install]'; echo 'no-cache-dir = true'; } > /etc/pip.conf
```

mattrobenolt:
@yosifkit I disagree. I don't think I'd want this explicitly disabled for all future pip invocations, even inside my layers. Maybe I want to leverage the pip caching for my local dev environment and I mount a volume into `/root/.cache`. I'd rather assume the default behavior of pip instead of figuring out later that it has been globally disabled and need to remove this file.

Thoughts?

yosifkit:
I concede the point.  Can you update your commit to get the other couple files:

```console
3.4/Dockerfile:30:      && pip3 install --upgrade pip==$PYTHON_PIP_VERSION \
3.4/slim/Dockerfile:49: && pip3 install --upgrade pip==$PYTHON_PIP_VERSION \
3.4/wheezy/Dockerfile:30:       && pip3 install --upgrade pip==$PYTHON_PIP_VERSION \
```

tianon:
The reason I'd personally prefer the configuration file is that users who
already do "RUN pip install ..." would be reaping the benefits of the
change without changing their Dockerfile, but the complaint about "least
surprise" definitely resonates.

Perhaps we start with this change and then upgrade to a file + image
documentation (noting that we've created this file) at some potential
future point?

mattrobenolt:
@yosifkit I updated to catch the `pip3` cases.

@tianon The reason I'm against the config file is that there is no way to opt out of the behavior AFAIK if it's in the file. So it's dictating a position down to users. I'd much rather document and lead by example than force a position down on users.

tianon:
It can still be opted out of in a file -- you can either remove the file or use an explicit `--cache-dir`, but not going away from the defaults in this case definitely resonates more with me.

tianon:
LGTM

yosifkit:
LGTM

PR #55, opened 2015-06-16
Add "libreadline-dev" to fix readline module usage

### LINKED ISSUE #53: readline module missing in python:2.7-slim
```
Python 2.7.10 (default, Jun  9 2015, 21:06:35) 
[GCC 4.9.2] on linux2
>>> import readline
ImportError: No module named readline
```

See https://docs.python.org/2/library/readline.html

Someone could argue that the python-slim only contains software relevant for servers, but sometimes it would be very helpful to get some debug interfaces working. For example iPython works way better when the `readline` module is available.

tianon:
Fixes #53

### DIFF
--- a/2.7/slim/Dockerfile
+++ b/2.7/slim/Dockerfile
@@ -28,6 +28,7 @@ RUN set -x \
 		libbz2-dev \
 		libc6-dev \
 		libncurses-dev \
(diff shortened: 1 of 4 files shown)


yosifkit:
LGTM

PR #65, opened 2015-11-13
Added Python 2.7 with Ubuntu Wily

fishgills:
I'm working on a project that requires some newer versions of libraries located in Ubuntu Wily. Thought I would try to help out.

### DIFF
--- a/2.7/wily/Dockerfile
+++ b/2.7/wily/Dockerfile
@@ -1,0 +1,41 @@
+FROM buildpack-deps:wily
+
+# remove several traces of debian python
+RUN apt-get purge -y python.*
+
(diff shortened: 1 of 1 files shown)


tianon:
Sorry for the delay -- honestly not sure how I feel about an official wily variant.  If we were to add this, why wouldn't we add every suite combination (which is a little bit insane for sure)? ie, what if anything makes "wily" a special case?

I can appreciate the need for newer libraries, but I don't know that there's a good solution for reconciling that here which doesn't lead down the path of having an insane number of variants. :disappointed:

mattrobenolt:
I'm +1 on closing this as well, especially considering _if anything_, there's new LTS.

PR #70, opened 2015-12-15
Add "all"

hobarrera:
This is a new image, which contains all supported python versions.  
The intention is for this image to be used in scenarios like running tests (in my particular case, via [GitLab CI](http://doc.gitlab.com/ci/docker/using_docker_build.html)).

The code is actually 100% based on existing `Dockerfile`s.

### DIFF
--- a/all/Dockerfile
+++ b/all/Dockerfile
@@ -1,0 +1,114 @@
+FROM buildpack-deps:jessie
+
+# remove several traces of debian python
+RUN apt-get purge -y python.*
+
+# http://bugs.python.org/issue19846
+# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
+ENV LANG C.UTF-8
+
+# gpg: key 18ADD4FF: public key "Benjamin Peterson <benjamin@python.org>" imported (for py27)
+RUN gpg --keyserver ha.pool.sks-keyservers.net --recv-keys C01E1CAD5EA2C4F0B8E3571504C367C218ADD4FF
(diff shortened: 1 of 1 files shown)


Starefossen:
Shouldn't testing be done in different containers in order to ensure that individual test runs does not interfere with each other?

hobarrera:
Python testing tools like [tox](https://testrun.org/tox/latest/) run test suites with different python versions, using each one at a time - while it's possible to configure a different container for each, it adds a configuration overhead, which then results in a runtime overhead with no real benefit.

hobarrera:
bump?

tianon:
I'm personally -1 to this as an official variant, especially since this isn't a configuration that upstream is especially keen to recommend or endorse (even if they do support it).  I think using something like https://github.com/yyuu/pyenv to replicate this kind of configuration inside the container is probably going to be easier to manage in the long run for folks who need more than one version in the same container.

hobarrera:
@tianon: Since my main target usage is CI, `pyenv` doesn't really help much; I'd need to start with a blank container and install everything onto is *on each run*. The isolate-python versions problem is already tackled by `tox`, a very common python test-suite runner.

Installing all python versions on each run is exactly what I avoid by using the container proposed in this PR.

tianon:
Right, I agree that an image like this _is_ useful, but I disagree that it
makes sense to include as an official Python variant, given that it's not a
standard configuration that upstream recommends.  My point was that it
should be reasonably easy to build an image that does this using existing
tools without adding an official tag here that does it, and that you're
going to have a better, more well-supported experience by doing so using
tools that are already designed to do this (and do it reasonably well).

PR #73, opened 2015-12-24
Add "-f" to all "curl" invocations

tianon:


tianon:
Nice, only failure is Python 3.5 + Hy, which is known (see also https://github.com/hylang/hy/pull/1006 for attempts to get the fix for that into a release).

PR #74, opened 2015-12-29
[RFC] Add 3.5/alpine variant

ncopa:
This is a continuation of #71 which was deleted by mistake.

I have made incremental changes for displaying the progress. I can merge the patches into a single before we merge it into docker-library if wanted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ services: docker
 env:
   - VERSION=3.5 VARIANT=
   - VERSION=3.5 VARIANT=slim
+  - VERSION=3.5 VARIANT=alpine
   - VERSION=3.4 VARIANT=
   - VERSION=3.4 VARIANT=slim
   - VERSION=3.4 VARIANT=wheezy
--- a/3.5/Dockerfile
+++ b/3.5/Dockerfile
@@ -16,12 +16,13 @@ ENV PYTHON_VERSION 3.5.1
 ENV PYTHON_PIP_VERSION 7.1.2
 
 RUN set -x \
-	&& mkdir -p /usr/src/python \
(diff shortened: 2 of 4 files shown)


ncopa:
Seems that the tests are failing due to missing bash and readline. Those were intentionally excluded so I guess we need to fix the test suite.

tianon:
Figured I'd just PR my comments so they're easier to discuss and/or incorporate: https://github.com/ncopa/python/pull/1 :smile:

tianon:
I've restarted the CI job now that the test is fixed -- this LGTM (going to wait for CI to come back green though)

tianon:
Oh right, readline.  We'll need to do something with that test, but I don't think this is worth holding up on account of that. :+1:

tianon:
Hmm, it _does_ only add ~2.75 MB to install.  Still worth considering just adding it, since it's a core Python module -- does installing it after the fact enable the core module, or does it have to be built in during Python source compilation time?  If it's hard or impossible to add after the fact, then I think we need to just include it unilaterally, but if we can add it afterwards without too much trouble, then we're fine and need to fix the test.

tianon:
```console
$ docker run -it --rm python:alpine sh
/ # apk add --no-cache readline
fetch http://dl-4.alpinelinux.org/alpine/v3.3/main/x86_64/APKINDEX.tar.gz
fetch http://dl-4.alpinelinux.org/alpine/v3.3/community/x86_64/APKINDEX.tar.gz
(1/1) Installing readline (6.3.008-r4)
Executing busybox-1.24.1-r7.trigger
OK: 13 MiB in 18 packages
/ # python
Python 3.5.1 (default, Dec 31 2015, 22:51:08) 
[GCC 5.3.0] on linux
>>> import readline
ImportError: No module named 'readline'
>>> 
```
:disappointed: :cry:

BUT!  The silver lining is that we already have the deps of `readline` installed, so it only takes up an extra ~638.9 kB. :+1: :metal:

PR #82, opened 2016-01-14
Apply the 3.5 alpine variant changes across the board for 3.x (now that they've settled)

tianon:


tianon:
Aww, looks like 3.2 and Alpine aren't great friends: https://travis-ci.org/docker-library/python/jobs/102478439

3.2 is EOL next month though, so I've just removed the Alpine variant (all other builds are currently green in https://travis-ci.org/docker-library/python/builds/102478418, so feel free to merge when happy without waiting for Travis again).  I don't think 3.2 + Alpine is really worth a lot of time debugging given that it's EOL next month anyhow.

yosifkit:
LGTM

PR #87, opened 2016-01-26
Remove 3.2 (EOL next month)

### LINKED ISSUE #84: Remove/freeze python 3.2
Lots of python libraries are dropping support for python 3.2.

Specifically, `virtualenv` 14.0.0 (https://virtualenv.pypa.io/en/latest/changes.html) and `pip` 8.0.0 (https://pip.pypa.io/en/stable/news/).

`urllib3` has dropped 3.2: https://github.com/shazow/urllib3/pull/787 This means that `requests` (https://github.com/kennethreitz/requests) will be dropping 3.2 since it depends on `urllib3`.

So given that this is the trend, I propose we either remove 3.2 from here, or explicitly lock it so dependencies don't get updated like in #83.

tianon:
https://www.python.org/dev/peps/pep-0392/#lifespan

Closes #84 (see discussion there for more context)

Thanks @mattrobenolt and @alex :+1:

(Also worth noting that Pip has officially been throwing a warning on Python 3.2 for quite some time now too.)


yosifkit:
LGTM

PR #90, opened 2016-02-26
Add missing build dependency for lzma module

### LINKED ISSUE #86: lzma module was not found in python:3-slim
When I try to import the module [lzma](https://docs.python.org/3/library/lzma.html):

```
Python 3.5.1 (default, Jan 13 2016, 18:56:49) 
[GCC 4.9.2] on linux
>>> import lzma
  File "/usr/local/lib/python3.5/lzma.py", line 26, in <module>
    from _lzma import *
ImportError: No module named '_lzma'
```
It seems that the library ```liblzma-dev``` is needed to build this module. See  [setup.py] (https://github.com/python/cpython/blob/master/setup.py#L1418) script.

ushuz:
Build dependency for `lzma` module in Python 3.3+ is missing on slim (`liblzma-dev`) and alpine (`xz-dev`) images, as @yosifkit mentioned in [#86](https://github.com/docker-library/python/issues/86#issuecomment-174711451).

This will close #86.

### DIFF
--- a/3.3/alpine/Dockerfile
+++ b/3.3/alpine/Dockerfile
@@ -35,6 +35,7 @@ RUN set -ex \
 		pax-utils \
 		readline-dev \
 		sqlite-dev \
+		xz-dev \
 		zlib-dev \
 	&& cd /usr/src/python \
 	&& ./configure --enable-shared --enable-unicode=ucs4 \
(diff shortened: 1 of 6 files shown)


yosifkit:
LGTM

tianon:
Added https://github.com/docker-library/official-images/commit/f659f773b7713fab2bb02a4a26eefc9e002145ca, restarted Travis, and we're all green! :+1:

LGTM :heart:

PR #96, opened 2016-03-19
added python onbuild  for alphine 

mwaaas:


yosifkit:
Onbuild images really only work on the Debian based images since they are using [`buildpack-deps`](https://hub.docker.com/_/buildpack-deps/) which gives us many libraries so that most anything installed from `requirements.txt` just works.  The alpine images are meant to be super tiny so a user will need to install any necessary libraries or build tools before doing a `pip install`.  We really only recommend `onbuild` as a starting point where most users should probably use one of the other variations and be able to add in their own custom steps.

PR #99, opened 2016-03-28
Fix python-config symlink

### LINKED ISSUE #97: Wrong python-config symlink in Python 3.5 alpine
There is a typo.
```
ln -s python-config3 python-config
```` 
should be 
```
ln -s python3-config python-config
```
https://github.com/docker-library/python/blob/master/3.5/alpine/Dockerfile#L67

yosifkit:
Fixes #97. (thanks @ushuz for reporting it)


tianon:
LGTM

PR #103, opened 2016-04-19
Explicitly empty ~/.cache

mattrobenolt:
This shaves off ~1.1MB off the container builds. We do --no-cache-dir on
`pip install`, but turns out, this isn't enough. Stuff from within
`get-pip.py` ends up leveraging the http cache as well, so we should
just nuke it entirely.

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -37,7 +37,7 @@ RUN set -ex \
 		\( -type d -a -name test -o -name tests \) \
 		-o \( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
 		-exec rm -rf '{}' + \
-	&& rm -rf /usr/src/python
+	&& rm -rf /usr/src/python ~/.cache
 
 # install "virtualenv", since the vast majority of users of this image will want it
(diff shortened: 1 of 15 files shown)


yosifkit:
Oh pip... :disappointed:  LGTM

tianon:
Can we use `~/.cache` instead just to help insulate ourselves a tiny bit from `HOME` screwiness? :innocent:

mattrobenolt:
> Can we use ~/.cache instead just to help insulate ourselves a tiny bit from HOME screwiness? :innocent:

Is it possible that it'd be built while running under a different user? I'm not against the change, just curious why that scenario may happen.

mattrobenolt:
Updated to use `~` instead of `/root`

tianon:
Not today, but we've been bitten by that very bug before, so I'm just
trying to be extra careful -- if you don't have any cycles to spare to
update that, I can do the minor change post-merge too. :+1:

tianon:
You were faster than GitHub's email to comment gateway :laughing:

LGTM, thanks :+1:

PR #107, opened 2016-05-08
Add support for sqlite3 extension loading

disarticulate:
[Run-Time Loadable Extensions](https://www.sqlite.org/loadext.html)

I'm not sure what the universe is for needing this, but it allows various things like [Spatialite](http://www.gaia-gis.it/gaia-sins/) to function at run-time with extensions.

### DIFF
--- a/3.5/alpine/Dockerfile
+++ b/3.5/alpine/Dockerfile
@@ -40,7 +40,7 @@ RUN set -ex \
 		xz-dev \
 		zlib-dev \
 	&& cd /usr/src/python \
-	&& ./configure --enable-shared --enable-unicode=ucs4 \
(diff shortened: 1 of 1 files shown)


tianon:
It's enabled in Debian's python3.5 package (https://sources.debian.net/src/python3.5/3.5.1-12/debian/rules/#L318), so I'm +1.

I think it will definitely need to be added consistently to all the variants of 3.5, and likely should be considered for all the 3.x series.  Would this apply to 2.7 as well?

tianon:
Debian doesn't apply it on 2.7 (https://sources.debian.net/src/python2.7/2.7.11-8/debian/rules/#L305-L311), so it's likely not available that far back, but I think it does make sense for other versions in the 3.x series.

PR #108, opened 2016-05-09
Add libpq so that psycopg2 can be installed.

klynton:
libpq (https://pkgs.alpinelinux.org/packages?name=libpq) provides the pg_config binary which is required for psycopg2 to build.

### DIFF
--- a/3.5/alpine/Dockerfile
+++ b/3.5/alpine/Dockerfile
@@ -30,6 +30,7 @@ RUN set -ex \
 		bzip2-dev \
 		gcc \
 		libc-dev \
+		libpq \
 		linux-headers \
 		make \
 		ncurses-dev \


yosifkit:
The alpine images are meant to be as minimal as possible and installing any external dependencies for your application like `psycopg2` would require you to install any apk dependencies that you require as well.  Also, unless [python itself is linking](https://github.com/docker-library/python/blob/87cd3580b7ca112790522da08173efece18443fb/3.5/alpine/Dockerfile#L53-L60) to the `libpq` library it would just be removed on [line 61](https://github.com/docker-library/python/blob/87cd3580b7ca112790522da08173efece18443fb/3.5/alpine/Dockerfile#L61).

klynton:
Gotcha! Thanks for the clear explanation of how this works. I'll install it another way.

PR #116, opened 2016-06-13
Remove default installation of virtualenv

### LINKED ISSUE #115: Possibly incorrect assumption that most people want virtualenv
After inspecting the Dockerfile for 2.7.11:wheezy I noticed that virtualenv gets installed by default: https://github.com/docker-library/python/blob/9383f7d4d2f96068e8957651aa3588fee8b48f71/2.7/wheezy/Dockerfile

I don't need virtualenv as I'm using Docker to run a single service where docker already provides isolation of dependencies. I think most people won't actually need virtualenv but instead will use pip to install packages globally.

My suggestion is to remove the line to install virtualenv and reduce the image to be clean and unbiased.

samos123:
Closes #115

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -41,7 +41,4 @@ RUN set -ex \
 		\) -exec rm -rf '{}' + \
 	&& rm -rf /usr/src/python ~/.cache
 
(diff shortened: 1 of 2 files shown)


tianon:
Closing given discussion in #115. :+1:

Issue #118, opened 2016-06-22
Old pip version is still present


michael-k:
Although [pip 8.1.2](https://github.com/docker-library/python/blob/9383f7d4d2f96068e8957651aa3588fee8b48f71/3.5/Dockerfile#L16) is installed in the docker image `python:3.5`, pip 7.1.2 is still present.

pip 8.1.2 is installed and used:
```
$ docker run --rm -it python:3.5 pip --version
pip 8.1.2 from /usr/local/lib/python3.5/site-packages (python 3.5)
```

pip lists an old version of itself as outdated:
```
$ docker run --rm -it python:3.5 pip list --outdated                       
pip (7.1.2) - Latest: 8.1.2 [wheel]
setuptools (18.2) - Latest: 23.0.0 [wheel]
You are using pip version 7.1.2, however version 8.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

It sill lies around in site-packages:
```
$ docker run --rm -it python:3.5 ls /usr/local/lib/python3.5/site-packages/
README	__pycache__  _markerlib  easy_install.py  pip  pip-7.1.2.dist-info  pip-8.1.2.dist-info  pkg_resources	setuptools  setuptools-18.2.dist-info
```

Regards pypa/pip#3538

tianon:
Interesting -- I'm not sure this is really pypa/pip#3538, since our official image is built on AUFS (not overlay), and since we do the `pip upgrade` directly in the same layer that it was originally installed (specifically to avoid issues like this), but I could be mistaken.  Is there a flag we're missing on our `pip` line to make sure it cleans up older artifacts completely?

michael-k:
> I'm not sure this is really pypa/pip#3538

It is, just not because of overlay.

> Is there a flag we're missing on our pip line to make sure it cleans up older artifacts completely?

Good idea, it happens due to `--ignore-installed`
https://github.com/docker-library/python/blob/9383f7d4d2f96068e8957651aa3588fee8b48f71/3.5/Dockerfile#L34
I removed it locally, built the image and ran the commands again.  The errors are gone.

michael-k:
PR docker-library/python#100 fixes this.

tianon:
I don't remember why we added `--ignore-installed` :disappointed:

If it currently works without, that sounds great, and we can cross the bridge of why we added it previously if the problem it solved resurfaces. :+1:

[...]

PR #100, opened 2016-04-09
Prevent two versions of pip

kane-c:
During the build, pip is upgraded, but the old version is not replaced.

This results in the pip update message being shown whenever you use it:
```
You are using pip version 7.1.2, however version 8.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

Even though the latest version is used:
```
$ pip -V
pip 8.1.1 from /usr/local/lib/python3.5/site-packages (python 3.5)
```

I noticed the `--ignore-installed` flag was backported from the 3.5 Dockerfile but couldn't find any explanation for it. Removing it worked well for me though.

PR #119, opened 2016-06-27
Do not ignore installed when updating pip

michael-k:
This eliminates all occurrences of `--ignore-installed` when updating pip.

Fixes docker-library/python#118


michael-k:
Closed in favor of docker-library/python#100

PR #131, opened 2016-07-22
Use "dpkg-divert" instead of "purge" to avoid Debian's Python

### LINKED ISSUE #129: pkg-config, libfreetype6-dev missing
It looks like as of yesterday `pkg-config` and `libfreetype6-dev` (and presumably other packages) are no longer in `python:2.7` (and presumably other tags). Was this deliberate and should I just start installing these myself in my images?

tianon:
Fixes #129

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -1,7 +1,19 @@
 FROM buildpack-deps:jessie
 
-# remove several traces of debian python
-RUN apt-get purge -y python.*
+# divert many traces of Debian Python (so that they are not used by mistake)
+# https://bugs.debian.org/33263 :(
+RUN set -ex \
+	&& for bits in \
+#		/etc/python* \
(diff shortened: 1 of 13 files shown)


yosifkit:
LGTM

os:
This is going to cause issues for people who are using Supervisor as this is going to remove pre-installed version. You can't run Supervisor with Python 3 and this is removing the old one.

```
bash: /usr/bin/supervisord: /usr/bin/python: bad interpreter: No such file or directory
```

glynjackson:
What is the reason behind this commit? a whole set of issues arise as @os has commented.

mozbhearsum:
There's some background in https://github.com/docker-library/python/issues/132#issuecomment-235691116. tl;dr - running purge removes some pcre dev libraries, which cause some things (notably uwsgi) to compile without necessary features.

PR #136, opened 2016-08-04
Alpine: install ca-certificates

### LINKED ISSUE #109: ca-certificates installed in slim images but not in alpine images
The `ca-certificates` package is installed in the `slim` images but not in the `alpine` images. Without it, HTTPS doesn't work (and so pip fails with standard settings).

It's not really hard to do a `apk add ca-certificates` but it's a bit unintuitive that there is this difference between the `slim` and `alpine` images.

Related: #92

JayH5:
Fixes #109 and hopefully the many related issues.

3.5:
- Before: 86.67 MB
- After: 87.95 MB

2.7:
- Before: 70.07 MB
- After: 71.35 MB

(i.e. adds 1.28MB, uncompressed)


yosifkit:
Sure, why not. LGTM, cc @tianon

PR #137, opened 2016-08-05
Revert "dpkg-divert" and change PATH instead

### LINKED ISSUE #133: /usr/bin/python: bad interpreter: No such file or directory
there's a bug in the latest 2.7 version of the container
```

docker pull python:2.7
docker run -it --rm python:2.7 bash -c 'apt-get update && apt-get install -y supervisor && supervisord —help'
...
...
Processing triggers for systemd (215-17+deb8u4) ...
bash: /usr/bin/supervisord: /usr/bin/python: bad interpreter: No such file or directory
```

### LINKED ISSUE #134: Newest revision broke compatibility
PR #131 broke my build, not entirely sure why but probably related to other reports.

Smallest failing code:

    FROM python:3.5
    RUN apt-get update && apt-get install -y gconf2

yosifkit:
Reverting and changing #131 which was the fix for #129.  Also fix some spaces to tabs like the rest of the file.

Fix #133, fix #134.

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -47,9 +47,9 @@ RUN set -ex \
 	&& [ "$(pip list | awk -F '[ ()]+' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ] \
 	&& find /usr/local -depth \
 		\( \
-		    \( -type d -a -name test -o -name tests \) \
-		    -o \
-		    \( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
+			\( -type d -a -name test -o -name tests \) \
+			-o \
+			\( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
 		\) -exec rm -rf '{}' + \
 	&& apt-get purge -y --auto-remove $buildDeps \
 	&& rm -rf /usr/src/python ~/.cache
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -1,19 +1,7 @@
 FROM buildpack-deps:jessie
 
-# divert many traces of Debian Python (so that they are not used by mistake)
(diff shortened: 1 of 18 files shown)


JayH5:
FWIW, these images already have `/usr/local/bin` in their `PATH`s before `/usr/bin`. `debian:jessie`, `buildpack-deps:jessie` and `buildpackdeps-wheezy` all have a (very standard) `PATH` of `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`. But I guess you're wanting to make that explicit...

tianon:
@JayH5 indeed, the point is to be 100% explicit to ensure that by default, `python` resolves to what we've created/added no matter what happens :+1:

LGTM

darabos:
Would it make sense to add automated tests for this issue? Python is one of the most popular images: an issue like this has a fairly large impact.

yosifkit:
@darabos, that sounds like a great idea. Do you have an idea of what would be the best indicator?  Maybe just test that `/usr/bin/python --version` is `Python 2.7.9`?  I am not sure what we are trying to test for so I am unsure where the test needs to fail.

We do have tests that we run from travis that are also part of accepting it into official-images for building and pushing to the hub: https://github.com/docker-library/official-images/tree/master/test.

darabos:
Sorry, I'm just a clueless user with no deep understanding of the issue. We just experienced that `apt-get install -y openjdk-7-jre` was failing.

Thanks for the link, I never knew about those tests! It looks like [`debian-apt-get/container.sh`](https://github.com/docker-library/official-images/blob/master/test/tests/debian-apt-get/container.sh) already tests `apt-get install` with a toy package. Would it be crazy to test with a more substantial package, like `openjdk-7-jre`?

Just testing `/usr/bin/python --version` like you propose would definitely be faster and more reliable. :+1:

PR #121, opened 2016-06-28
Fix two versions of pip being installed, and add a basic test to ensure the pip version we request is the pip version we get

tianon:
Fixes #118 
Closes #100

PR #143, opened 2016-08-19
Python 3.6.0a4

JayH5:


JayH5:
For whatever reason pip isn't available as `pip` in 3.6.0a4, only as pip3, which is causing the image to fail to build 😢

tianon:
Aha!  We finally found the edge case that led to adding `--ignore-installed`! (#121, #118, #100)

Looks like Python 3.6.0a4 already comes with the latest version of `pip`, so it no-ops when we ask it to install, which means it doesn't create the `pip` symlink properly. :metal:

I _think_ `--force-reinstall` is the option we really want (instead of `--ignore-installed`) to ensure it cleans up references to the older versions properly (which is what that `pip list` line is verifying).  If you'd like to try taking that change on, you're welcome to, otherwise I'm happy to take over from here (and think we'd be fine to merge this PR as-is, since the version bump is correct even if the build is currently failing). :+1:

JayH5:
@tianon you can go ahead and take over 😄. `--force-reinstall` sounds about right (although maybe with a comment this time so that we remember why 😛)

PR #149, opened 2016-10-11
Update to 3.6.0b2

IlyaSkriblovsky:


lelit:
The build of 3.6.0b2 on hub.docker.com failed due to a network (temporary) failure.
Can you retrigger it please?

tianon:
@lelit not sure what you're referring to -- got a link to the failure you're seeing?

This is currently waiting on https://github.com/docker-library/official-images/pull/2242 to go live (which is being build-tested as we speak).

lelit:
Ok, no problem: yesterday, noticing the switch to b2, I checked the build log and it showed a network failure while downloading some library (IIRC, libtcl-dev).
Thank you!

PR #153, opened 2016-10-25
Add gdbm-dev to slim and alpine images

yosifkit:
Happy green tests from https://github.com/docker-library/official-images/pull/2284.

 Non-slim, non-alpine images should still fail until a new buildpack-deps is created.

### DIFF
--- a/2.7/slim/Dockerfile
+++ b/2.7/slim/Dockerfile
@@ -10,6 +10,7 @@ ENV LANG C.UTF-8
 # runtime dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		ca-certificates \
+		libgdbm3 \
 		libsqlite3-0 \
 		libssl1.0.0 \
 	&& rm -rf /var/lib/apt/lists/*
--- a/2.7/slim/Dockerfile
+++ b/2.7/slim/Dockerfile
@@ -26,6 +26,7 @@ RUN set -ex \
 		libbz2-dev \
 		libc6-dev \
 		libdb-dev \
+		libgdbm-dev \
 		libncurses-dev \
 		libreadline-dev \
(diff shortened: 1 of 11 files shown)


tianon:
```
testing python:3.6-slim
	'utc' [1/8]...passed
	'cve-2014--shellshock' [2/8]...passed
	'no-hard-coded-passwords' [3/8]...passed
	'override-cmd' [4/8]...passed
	'python-hy' [5/8]...skipping Hy test -- no workie on Python 3.6+
passed
	'python-imports' [6/8]...Traceback (most recent call last):
  File "./container.py", line 15, in <module>
    import dbm.gnu
  File "/usr/local/lib/python3.6/dbm/gnu.py", line 3, in <module>
    from _gdbm import *
ImportError: libgdbm.so.3: cannot open shared object file: No such file or directory
failed
	'python-pip-requests-ssl' [7/8]...passed
	'python-sqlite3' [8/8]...passed
```

yosifkit:
Hmmm, that is similar to how `tk` currently works; we add it for building but require the user to install the apt package of they want to use it.  Do we want to keep it that way?

It is "better" than the old error:
```
Python 3.6.0b2 (default, Oct 15 2016, 06:24:15) 
[GCC 4.9.2] on linux
>>> import dbm.gnu
  File "/usr/local/lib/python3.6/dbm/gnu.py", line 3, in <module>
    from _gdbm import *
ModuleNotFoundError: No module named '_gdbm'
```

yosifkit:
:green_heart: Green in all the right places!

tianon:
Yeah, keeping it around here seems more sane IMO than TK -- it's likely more common and should have a smaller footprint overall anyhow. :+1:

LGTM

Issue #151, opened 2016-10-20
Python Docker images are missing dbm.gnu support


sebk-666:
Hi,
I tested multiple of the official Python 3.x Docker images and they all are missing support for dbm.gnu.
While there is a `<lib>/dbm/gnu.py` file that imports `_gdbm`, the corresponding `_gdbm.cpython.*.so`library in `<lib>/lib-dynload/` is missing.

I tested on python:3.4, python:3.5, python:3.6, python:3.4-alpine - all with the same result.

```
# python3
Python 3.4.5 (default, Sep 24 2016, 05:29:30)
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dbm.gnu
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.4/dbm/gnu.py", line 3, in <module>
    from _gdbm import *
ImportError: No module named '_gdbm'
```

Sebastian

yosifkit:
Looks like we need `libgdbm-dev` and since we already are adding it to ruby images I think this makes it a candidate for addition to `buildpack-deps` so they can share the layer.  As a bonus it won't increase the image size by much:

```
After this operation, 167 kB of additional disk space will be used.
```

We'll still need to add the appropriate packages to slim and alpine variants here, but I've started a PR to add it to buildpack-deps. https://github.com/docker-library/buildpack-deps/pull/49

tianon:
Appears this was resolved with https://github.com/docker-library/buildpack-deps/pull/49 / https://github.com/docker-library/python/pull/153. :tada:

```console
$ docker pull python:3
3: Pulling from library/python
Digest: sha256:4748e00099548b67f250ee6f1210f1f5675113948ed0aae48e195ad6c8e12a32
Status: Image is up to date for python:3

$ docker run -it --rm python:3
Python 3.6.5 (default, Mar 31 2018, 01:15:58) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dbm.gnu
>>> 
```

PR #83, opened 2016-01-20
pip 8.0

mattrobenolt:
https://pip.pypa.io/en/stable/news/#release-notes

pip 8.0 introduces some backwards incompatibilities, so not even entirely sure if this is a great idea.

But on the otherhand, we have no way of versioning `pip` independently from `python` in these builds.

So we may end up with someone tagging to a very specific `python:2.7.11` which had pip 7.1.2, then tomorrow, gets new build with 8.0.0 and breaks their container.

Thoughts?

PR #154, opened 2016-11-04
Upgraded to pip 9.0.1

michael-k:


mattrobenolt:
fwiw, just weighing in my opinion, this change does bring backwards incompatibility, and we definitely don't have any way of tagging to indicate a version of pip. So a `python:2.7.12` pulled today is going to get 8.1.2, then pulling same image after this is merged is going to bring in 9.0.0 which may or may not break user applications.

Granted, I don't think that should hold us back from doing the upgrade, just something to keep in mind. I dealt with this when we upgraded to 8.0. :) See https://github.com/docker-library/python/pull/83 for discussion.

Definitely need to figure out why tests are failing though.

tianon:
Indeed, I'm wary of 9.0 given our experience with 8.0 (especially breaking our tests, which exposed a bug in `pip` itself). :cold_sweat:

mattrobenolt:
I'll have some time to look at this over the weekend and see if something obvious is standing out why tests would fail for this. 😭

tianon:
I think the tests might actually be failing due to https://github.com/docker-library/official-images/pull/2284 (docker-library/buildpack-deps#49, docker-library/python#151).

tianon:
Which doesn't really make sense, given that `buildpack-deps` was updated in https://github.com/docker-library/official-images/pull/2292 :disappointed:

mattrobenolt:
😨

michael-k:
All test that fail, fail on master too.  I'll guess it's best to fix the tests on master and then rebase.

michael-k:
Updated the commit to pip 9.0.1 and upgraded pip for the 3.6 image too (missed that before).

The tests pass now, I guess due to:
> `libgdbm-dev` is not needed since docker-library/buildpack-deps#49, but it looks like there was a hiccup in the signing process and so the new image is not showing up on the Docker Hub.

See https://github.com/docker-library/python/pull/156#issuecomment-258987457

If upgrading pip is so critical, maybe there should be images for the latest major version of pip and the one before.

michael-k:
rebased

tianon:
> If upgrading pip is so critical, maybe there should be images for the latest major version of pip and the one before.

I wouldn't say it's necessarily _critical_, it's just caused problems in the past.  I think we've got some pretty decent test coverage though, so I feel OK about updating this. :+1:

PR #156, opened 2016-11-07
Python 3.6 beta 3 + pip upgrade

askoretskiy:
Python 3.6 beta 3 is out, it's time to upgrade ))

Also with pip 9.0.1 it is time to move on.


askoretskiy:
I could apply the same changes to fix all base images (basically add libgdbm-dev to all containers)

yosifkit:
`libgdbm-dev` is not needed since https://github.com/docker-library/buildpack-deps/pull/49, but it looks like there was a hiccup in the signing process and so the new image is not showing up on the Docker Hub.  We'll push them again today.

PR #161, opened 2016-12-08
Update 3.6.0rc2 to 3.6.0

mike9005:
Makes the [12/7 release of 3.6.0](https://www.python.org/ftp/python/3.6.0/) available on Docker Hub

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -3,7 +3,6 @@ set -eu
 
 declare -A aliases=(
 	[3.6]='3 latest'
-	[3.5]='3 5'
 	[2.7]='2'
 )
 
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -2,8 +2,8 @@
 set -eu
 
 declare -A aliases=(
-	[3.6-rc]='rc'
-	[3.5]='3 latest'
+	[3.6]='3 latest'
+	[3.5]='3 5'
 	[2.7]='2'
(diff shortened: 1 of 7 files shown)


yosifkit:
That is strange; the artifacts aren't on `www.python.org/ftp/python/3.6.0/` yet: https://travis-ci.org/docker-library/python/jobs/182346726#L406-L411.

mike9005:
@yosifkit - yeah I noticed that - on closer inspection, it's code frozen and all other things being equal, that `rc1` release will be converted to `3.6.0` next Friday. Looks like I jumped the gun here - will re-run the tests when the FTP is updated 👍

generate-stackbrew-library.sh
```diff
@@ -2,8 +2,7 @@
 set -eu
 
 declare -A aliases=(
-	[3.6-rc]='rc'
```
tianon: When you come back around to updating this, can you change this line to be `[3.7-rc]='rc'` instead so we're ready for the next version's RC cycles instead of removing this line entirely? :pray: (it's harmless to have an unused line here)

mike9005: Sure thing - done.

msurdi:
It looks like the final package release has been already uploaded. Anyone here that can retry this? Thanks!

mike9005:
@tianon @yosifkit - rebased and passing CI - ought to be good to go, PTAL

tianon:
:metal:

PR #172, opened 2017-01-18
Update Python 3.5 to version 3.5.3.

dbaxa:
Signed-off-by: David Black <dblack@atlassian.com>

### DIFF
--- a/3.5/Dockerfile
+++ b/3.5/Dockerfile
@@ -20,7 +20,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GPG_KEY 97FC712E4C024BBEA48A61ED3A5CA953F73C700D
-ENV PYTHON_VERSION 3.5.2
+ENV PYTHON_VERSION 3.5.3
 
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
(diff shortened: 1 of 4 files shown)


tianon:
We seem to be having some issues where https://www.python.org/ftp/python/3.5.3/ is being over-aggressively cached in some places (my office here used to be affected, but appears to be updated now, while our CI server is still affected).

The end result of that over-aggressive caching is that if we merge this now, @docker-library-bot will turn right around and undo the change in a few hours until that cache clears and it sees the update. :disappointed:

tianon:
Looks like that cache is finally cleared as of today :+1:

jgoclawski:
@tianon Thanks for quick reaction! Can we get the image to show on the Docker Hub - https://hub.docker.com/r/library/python/ ? Python 3.5.3 is still not available there.

yosifkit:
We just made a PR to official images https://github.com/docker-library/official-images/issues/2571

> See a change merged here that doesn't show up on the Docker Hub yet? Check [the "library/python" manifest file in the docker-library/official-images repo](https://github.com/docker-library/official-images/blob/master/library/python), especially [PRs with the "library/python" label on that repo](https://github.com/docker-library/official-images/labels/library%2Fpython). For more information about the official images process, see the [docker-library/official-images readme](https://github.com/docker-library/official-images/blob/master/README.md).
>
> \- [README.md](https://github.com/docker-library/python/tree/4733c43dc67ab8515430ece69de6cdab27126d16#about-this-repo)

PR #181, opened 2017-03-22
Provide syntax for building images via URL before they land in DockerHub

cclauss:
This syntax is ~NOT~ correct (_fixed in second commit_) but I could not figure out the right syntax and I am sure others will have the same problem as me.


yosifkit:
I don't really want to have to update the README to point out how to build each new version (since the URL is for a specific release). I believe most users will know how to `git clone` and then can `docker build` within the version that they want.  Besides most releases are PR'ed and pushed to the Docker Hub within a week or less.

Issue #170, opened 2017-01-10
Update setuptools in python3.6


mackeyja92:
The setuptools included in the python3.6 version is broken and needs to be updated. Here is the specific issue https://github.com/pypa/setuptools/issues/866

A quick workaround to overwrite the included one is to add the following run command to your Dockerfile.

```
RUN pip install -U setup tools
```

yosifkit:
Looks like python should offer a new release with a working `setuptools`. :wink::wink:

But I guess we'll have to work around it and fix it similar to how we [keep pip up to date](https://github.com/docker-library/python/blob/0380e847a80e636b58d7be5778ffa420bd1c99d2/3.6/Dockerfile#L59-L65). :nauseated_face:

jouve:
Hi,

I also encountered the error while while installing python-memcached:
```
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-awdnpsmf/python-memcached/setup.py", line 8, in <module>
        version=get_module_constant('memcache', '__version__'),
      File "/usr/local/lib/python3.6/site-packages/setuptools/depends.py", line 164, in get_module_constant
        return extract_constant(code, symbol, default)
      File "/usr/local/lib/python3.6/site-packages/setuptools/depends.py", line 197, in extract_constant
        const = code.co_consts[arg]
    IndexError: tuple index out of range
```

@yosifkit  Would you accept a PR adding 
```pip install -U setuptools```
to the Dockerfile ?

yosifkit:
Given that the current version of setuptools is [28.8.8](https://github.com/pypa/setuptools/releases/tag/v28.8.0), released in November 2016, and the newest is 34.3.3, released a day ago, we definitely need a better way to keep it up to date.  I had thought that the 3.6.1 bump would've included a newer setuptools :cry:

Just adding the `pip install -U setuptools` would not be enough to trigger a rebuild when a new version of setuptools comes out, so we'll have to make sure that a version is specified like pip. Which means that we'll also need to change `update.sh` to grab the setuptools version and bump all the files.

```diff
# line 62 https://github.com/docker-library/python/blob/32e920eb13714a9aeff2e016fb467901222d17b5/3.6/Dockerfile#L62
- 	&& pip3 install --no-cache-dir --upgrade --force-reinstall "pip==$PYTHON_PIP_VERSION" \
+	&& pip3 install --no-cache-dir --upgrade --force-reinstall \
[...]

PR #186, opened 2017-03-31
Always run pip installer to get implicit deps

### LINKED ISSUE #141: Wheel is not included in Python 3
The wheel package is included in Python 2 but is not in Python 3:

```
docker run --rm python:2 bash -c "pip list | grep wheel"
wheel (0.29.0)

docker run --rm python bash -c "pip list | grep wheel"
```

This results in pip not building wheels in Python 3, while it does build them in Python 2.

mal:
Unconditionally executing the contents of [this `if` block](https://github.com/docker-library/python/blob/32e920eb13714a9aeff2e016fb467901222d17b5/Dockerfile-alpine.template#L66) results in the installation of `wheel` (consistent with <= 3.3 images) without over-zealous overhead:

```
[...] # end of make install output
Successfully installed pip-9.0.1 setuptools-28.8.0
+ wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py
Connecting to bootstrap.pypa.io (151.101.0.175:443)
get-pip.py           100% |*******************************|  1558k  0:00:00 ETA
+ python3 /tmp/get-pip.py pip==9.0.1
Requirement already up-to-date: pip==9.0.1 in /usr/local/lib/python3.6/site-packages
Collecting wheel
  Downloading wheel-0.29.0-py2.py3-none-any.whl (66kB)
Installing collected packages: wheel
Successfully installed wheel-0.29.0
+ rm /tmp/get-pip.py
[...]
```

Fixes #141, #170

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -18,6 +18,8 @@ ENV PYTHON_VERSION 2.7.13
 
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
 ENV PYTHON_PIP_VERSION 9.0.1
+ENV PYTHON_SETUPTOOLS_VERSION 35.0.1
+ENV PYTHON_WHEEL_VERSION 0.29.0
 
 RUN set -ex \
 	&& buildDeps=' \
@@ -50,7 +52,10 @@ RUN set -ex \
 # we use "--force-reinstall" for the case where the version of pip we're trying to install is the same as the version bundled with Python
 # ("Requirement already up-to-date: pip==8.1.2 in /usr/local/lib/python3.6/site-packages")
 # https://github.com/docker-library/python/pull/143#issuecomment-241032683
-	&& pip install --no-cache-dir --upgrade --force-reinstall "pip==$PYTHON_PIP_VERSION" \
+	&& pip install --no-cache-dir --upgrade --force-reinstall \
+		"pip==$PYTHON_PIP_VERSION" \
+		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		"wheel==$PYTHON_WHEEL_VERSION" \
 	\
 	&& find /usr/local -depth \
 		\( \
--- a/2.7/alpine/Dockerfile
+++ b/2.7/alpine/Dockerfile
@@ -16,6 +16,8 @@ ENV PYTHON_VERSION 2.7.13
 
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
 ENV PYTHON_PIP_VERSION 9.0.1
+ENV PYTHON_SETUPTOOLS_VERSION 35.0.1
+ENV PYTHON_WHEEL_VERSION 0.29.0
 
 RUN set -ex \
 	&& apk add --no-cache --virtual .fetch-deps \
@@ -67,7 +69,10 @@ RUN set -ex \
 # we use "--force-reinstall" for the case where the version of pip we're trying to install is the same as the version bundled with Python
 # ("Requirement already up-to-date: pip==8.1.2 in /usr/local/lib/python3.6/site-packages")
 # https://github.com/docker-library/python/pull/143#issuecomment-241032683
-	&& pip install --no-cache-dir --upgrade --force-reinstall "pip==$PYTHON_PIP_VERSION" \
+	&& pip install --no-cache-dir --upgrade --force-reinstall \
+		"pip==$PYTHON_PIP_VERSION" \
+		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		"wheel==$PYTHON_WHEEL_VERSION" \
 	\
 	&& find /usr/local -depth \
 		\( \
--- a/2.7/slim/Dockerfile
+++ b/2.7/slim/Dockerfile
@@ -20,6 +20,8 @@ ENV PYTHON_VERSION 2.7.13
 
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
 ENV PYTHON_PIP_VERSION 9.0.1
+ENV PYTHON_SETUPTOOLS_VERSION 35.0.1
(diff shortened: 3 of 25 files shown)


tianon:
I'm a little bit confused about what this does that just `pip3 install --no-cache-dir --upgrade --force-reinstall pip==9.0.1` is not doing (which is invoked later).

@dstufft would you be willing to give us some advice on whether it's sane / smart to always invoke `get-pip.py`, regardless of whether our Python install already includes a `pip3` which we can use for bootstrapping via `--force-reinstall`?  Any (other) suggestions for improvement? :pray: :innocent: :heart: :heart:

mal:
It installs `wheel` which is not present in any of the existing `> 3.3` images. If there's a better way of achieving the same, all the better 😄

dstufft:
I'm on my phone but a quick glance suggests the impetus for this change is that get-pip.py installs pip/setuptools/wheel but ensurepip only installs pip/setuptools. Expanding your later command to also include setuptools and wheel should solve this also I think. 

I will take a closer look when I get home.

tianon:
Oh, we also have https://github.com/docker-library/python/issues/170, where folks are suggesting that we need newer `setuptools`, so perhaps simply keeping `setuptools` and `wheel` up-to-date in the same way we handle `pip` would make sense? :smile:

(Definitely appreciate the input :heart: :heart: :heart:)

mal:
Just did a quick PoC of the suggested approach, following along with how pip version management is done.

There's a version difference between the `wheel` installed by `pip`'s `get-pip.py` and the `wheel` installed as a result of version grabbing from `pypi`: `wheel-0.29.0` vs `wheel-0.30.0a0`. I'm unsure if that's an issue but worth noting.

Happy to update this PR when we know which direction we'd like to take this 😄

dstufft:
That'll be because pip excludes pre-releases by default.

tianon:
Ah, `get-pip.py` includes the following:
```python
    # Add any implicit installations to the end of our args
    if implicit_pip:
        args += ["pip"]
    if implicit_setuptools:
        args += ["setuptools"]
    if implicit_wheel:
        args += ["wheel"]
```
(which just installs the latest release of each)

Seems roughly sane (and I think the way we fetch the latest `pip` version in `update.sh` accounts for that already, and if not, should be fixed and then copied for doing `setuptools` and `wheel` also).

mal:
In that case I think the current way of getting the package versions might need a tweak, because as I pointed out above, right now it happily grabs pre-release versions.

```sh
# curl -fsSL 'https://pypi.python.org/pypi/pip/json' | awk -F '"' '$2 == "version" { print $4 }'
9.0.1
# curl -fsSL 'https://pypi.python.org/pypi/setuptools/json' | awk -F '"' '$2 == "version" { print $4 }'
34.3.3
# curl -fsSL 'https://pypi.python.org/pypi/wheel/json' | awk -F '"' '$2 == "version" { print $4 }'
0.30.0a0
```

A look at the [docs](https://wiki.python.org/moin/PyPIJSON) claims that it will fetch the stable version by default but that does not appear to be happening:

> This retrieves information about the latest stable release

In the meantime, I'll update this PR with the version that grabs pre-releases and we can go forward from there.

mal:
I'll squash this down once we're happy with it 😉

mal:
Is it permissable to use tools such as `jq` when processing the `https://pypi.python.org/pypi/wheel/json` style URLs, or are we limited to using common system tools?

tianon:
Yeah, `jq` is one that's permissible (we use it in other places, so it's safe to assume it being installed).  I was thinking of replacing the `awk` with `jq` since that'd be more reliable anyhow. :smile:

I think the main issue I see is that the alpha is listed as the actual "latest release" for `wheel` by PyPI, so I'm not sure how we're expected to know that 0.29.0 is what we ought to be consuming unless we implement the PEP for Python version number sorting ourselves. :disappointed:

dstufft:
Try ``https://pypi.org/pypi/wheel/json``.

mal:
Aha! Perfect! I'll get a commit going, thanks 😁

tianon:
You're a hero, @dstufft :heart: (thanks for all you do! :+1:)

dstufft:
I feel obligated to mention that URL is technically not "production" which mostly just means there's no alerting if it goes down, but that should be fairly rare to get downtime of that URL that actually makes it unavailable.

tianon:
Yeah, I figured as much when I checked out https://pypi.org (and saw the warning), but it's only going to be used from our `update.sh` script that's used for updating the version numbers, not during builds, so I'm +1 on using it since it solves our issues. :smile: :heart: (especially since now we can claim to be guinea pigs helping test pypi.org! :innocent:)

mal:
Anything more to do on this one @tianon?

tianon:
I took the liberty of rebasing this on https://github.com/docker-library/python/pull/187 for you, and added a commit (60d03bd670bb9956b9c62a1004640e88669c301c) which both fixes our `pip` update code on Windows, and applies this same change there. :+1: :metal:

JayH5:
I'm a bit concerned about tracking the `setuptools` version as it moves very fast which would lead to these images being rebuilt very often. For example, there's already version 35.0.2.

https://github.com/pypa/setuptools/blame/master/CHANGES.rst

tianon:
Hmm, good point @JayH5.  @yosifkit and I took a look at Python's `./configure` options, and we're considering using `--without-ensurepip` to have Python no longer install the embedded `pip` at all, and just use `get-pip.py` unilaterally, which means we can even move the `get-pip.py` into a separate layer and get even better cache (especially for minor `pip` bumps).  :sweat_smile:

(Then, since we'll be using `get-pip.py` unconditionally, we'll always get `setuptools` and `wheel`, and they'll be the "latest release at the time of our installation of `pip`", which also solves the problem of "when do we bump them?" -- they'll be naturally bumped any time Python, `pip`, or our base image are. :+1:)

tianon:
(Which ironically enough is basically exactly what @mal suggested originally -- sorry @mal :see_no_evil: :heart: :heart:)

PR #187, opened 2017-03-31
Fix pip list deprecation warning

mal:
While working on another PR I noticed the following deprecation warning in the output:

```
[...]
Successfully installed pip-9.0.1
+ awk -F [ ()]+ $1 == "pip" { print $2; exit }
+ tac
+ tac
+ pip list
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
+ [ 9.0.1 = 9.0.1 ]
[...]
```

To pre-empt this change I've moved the check over to using `pip freeze`. I think `pip freeze --all` is a better candidate for this check as it not only has an easier to parse explicitly defined format, but is also less likely to change.

---

As an aside, I do not understand the purpose of the `|tac|tac|` pipe in the affected line (maybe buffering?), but have left it in place as history seems to suggest it was intentionally added during the move to templates, unfortunately no explicit reasoning is given. In some cursory testing it appeared to make no difference, but if this is some sort of intentional trick I would very much like to understand it, if anyone is willing to share their insight. 😄

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -51,9 +51,6 @@ RUN set -ex \
 # ("Requirement already up-to-date: pip==8.1.2 in /usr/local/lib/python3.6/site-packages")
 # https://github.com/docker-library/python/pull/143#issuecomment-241032683
 	&& pip install --no-cache-dir --upgrade --force-reinstall "pip==$PYTHON_PIP_VERSION" \
-# then we use "pip list" to ensure we don't have more than one pip version installed
-# https://github.com/docker-library/python/pull/100
-	&& [ "$(pip list |tac|tac| awk -F '[ ()]+' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ] \
 	\
 	&& find /usr/local -depth \
 		\( \
--- a/2.7/alpine/Dockerfile
+++ b/2.7/alpine/Dockerfile
@@ -68,9 +68,6 @@ RUN set -ex \
 # ("Requirement already up-to-date: pip==8.1.2 in /usr/local/lib/python3.6/site-packages")
 # https://github.com/docker-library/python/pull/143#issuecomment-241032683
 	&& pip install --no-cache-dir --upgrade --force-reinstall "pip==$PYTHON_PIP_VERSION" \
-# then we use "pip list" to ensure we don't have more than one pip version installed
-# https://github.com/docker-library/python/pull/100
-	&& [ "$(pip list |tac|tac| awk -F '[ ()]+' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ] \
 	\
 	&& find /usr/local -depth \
 		\( \
--- a/2.7/slim/Dockerfile
+++ b/2.7/slim/Dockerfile
@@ -66,9 +66,6 @@ RUN set -ex \
 # ("Requirement already up-to-date: pip==8.1.2 in /usr/local/lib/python3.6/site-packages")
 # https://github.com/docker-library/python/pull/143#issuecomment-241032683
 	&& pip install --no-cache-dir --upgrade --force-reinstall "pip==$PYTHON_PIP_VERSION" \
-# then we use "pip list" to ensure we don't have more than one pip version installed
-# https://github.com/docker-library/python/pull/100
-	&& [ "$(pip list |tac|tac| awk -F '[ ()]+' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ] \
 	\
 	&& find /usr/local -depth \
 		\( \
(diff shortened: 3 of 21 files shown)


2.7/Dockerfile
```diff
@@ -44,16 +44,16 @@ RUN set -ex \
 	&& make install \
 	&& ldconfig \
 	\
-		&& wget -O /tmp/get-pip.py 'https://bootstrap.pypa.io/get-pip.py' \
-		&& python2 /tmp/get-pip.py "pip==$PYTHON_PIP_VERSION" \
-		&& rm /tmp/get-pip.py \
+	&& wget -O /tmp/get-pip.py 'https://bootstrap.pypa.io/get-pip.py' \
+	&& python2 /tmp/get-pip.py "pip==$PYTHON_PIP_VERSION" \
+	&& rm /tmp/get-pip.py \
```
tianon: This looks like it was unintentionally added here -- these are kept indented so that diff between this version and the other versions is smaller (no whitespace changes, only removed lines, essentially).

2.7/Dockerfile
```diff
@@ -44,16 +44,16 @@ RUN set -ex \
...
 # we use "--force-reinstall" for the case where the version of pip we're trying to install is the same as the version bundled with Python
 # ("Requirement already up-to-date: pip==8.1.2 in /usr/local/lib/python3.6/site-packages")
 # https://github.com/docker-library/python/pull/143#issuecomment-241032683
 	&& pip install --no-cache-dir --upgrade --force-reinstall "pip==$PYTHON_PIP_VERSION" \
-# then we use "pip list" to ensure we don't have more than one pip version installed
+# then we use "pip freeze" to ensure we don't have more than one pip version installed
 # https://github.com/docker-library/python/pull/100
-	&& [ "$(pip list |tac|tac| awk -F '[ ()]+' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ] \
+	&& [ "$(pip freeze --all |tac|tac| awk -F '==' '$1 == "pip" { print $2; exit }')" = "$PYTHON_PIP_VERSION" ] \
```
tianon: Very nice -- `pip freeze` is indeed a much better solution. :+1:

The `|tac|tac|` is a cheeky trick to ensure the entire output of `pip freeze` is read completely -- some applications get really grumpy when we do `exit` from `awk` too quickly and standard input is then closed (causing `pip freeze` to have standard output closed).  It's easier to illustrate with `curl`+`head`:

```console
$ curl -fsSL 'https://google.com' | head -1 > /dev/null
curl: (23) Failed writing body (867 != 1339)
```

mal: Aha! Thanks for the clarification 👍

mal:
Removed the 2.7 whitespace commit. I also had a brief look at replicating the issue in #100 but wasn't able to, due to `ensurepip` currently installing `9.0.1`, the same version specified in the Dockerfile.

mal:
I just built a `3.5` image from e327311c1ec5400ce6bb7cc0b2ff34326ea659ca (commit prior to the fix in #100) and was unable to reproduce the issue at all:

```sh
[... snip ...]
Successfully installed pip-7.1.2 setuptools-18.2
+ ldconfig
+ pip3 install --no-cache-dir --upgrade --ignore-installed pip==8.1.1
Collecting pip==8.1.1
  Downloading pip-8.1.1-py2.py3-none-any.whl (1.2MB)
Installing collected packages: pip
Successfully installed pip-8.1.1
[... snip ...]
```

```sh
# pip freeze
You are using pip version 8.1.1, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
# pip3 freeze
You are using pip version 8.1.1, however version 9.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

I was expecting to see `7.1.2, however version 9.0.1 is available.` in order for it to be consistent with the original report, but seeing 8.1.1 means the correct version for that image is both installed and recognised.

Digging a little deeper I found references to `7.1.2` but none that actively surfaced in the manner described in the report:

```
# pip3 -V      
pip 8.1.1 from /usr/local/lib/python3.5/site-packages (python 3.5)
# ls -l /usr/local/lib/python3.5/site-packages
total 40
-rw-r--r--  1 root staff  119 Apr  7 20:49 README
drwxr-sr-x  2 root staff 4096 Apr  7 20:49 __pycache__
drwxr-sr-x  3 root staff 4096 Apr  7 20:49 _markerlib
-rw-r--r--  1 root staff  126 Apr  7 20:49 easy_install.py
drwxr-sr-x 20 root staff 4096 Apr  7 20:50 pip
drwxr-sr-x  2 root staff 4096 Apr  7 20:49 pip-7.1.2.dist-info
drwxr-sr-x  2 root staff 4096 Apr  7 20:49 pip-8.1.1.dist-info
drwxr-sr-x  4 root staff 4096 Apr  7 20:49 pkg_resources
drwxr-sr-x  4 root staff 4096 Apr  7 20:49 setuptools
drwxr-sr-x  2 root staff 4096 Apr  7 20:49 setuptools-18.2.dist-info
```

At this point I'm at a bit of a loss. The proposed change makes sense in the context of updating for deprecation, but it seems like the reason for this check being present at all could use some further investigation, probably from someone with more familiarity with pip and python module internals than me I'm afraid 😞

tianon:
Hmm, OK -- maybe then we should simply remove the check entirely and revisit if the problem comes back.  Sound sane?

mal:
Think so, yes. If nothing else the `--force-install` option on the line above wasn't present at the time this existed as an issue, and I'd expect that to prevent it recurring. I'll update this PR 👍

Issue #160, opened 2016-11-27
Consider adding '--enable-optimizations' configure argument


hannosch:
Python has recently added a new configure argument called ``--enable-optimizations``, one of the relevant issues is https://bugs.python.org/issue26359.

The new argument will turn on compile time optimizations relevant to production builds. Currently this turns on profile guided optimizations (PGO). PGO builds will take a lot longer (possible 20-40 minutes) to produce, but the resulting Python binary is about 10% faster at executing Python code (according to [Python's new benchmark suite](https://github.com/python/performance)).

The new configure flag is available in the recently released Python 3.6.0b4 and was backported to the 3.5 and 2.7 branches, so it should become available in Python 2.7.13 and 3.5.3.

In the future the Python developers may decide to turn on further optimizations based on this argument, for example link-time optimizations (LTO), though they haven't worked out all the bugs for that one yet.

I think the docker images should use this argument, to benefit from the added optimizations.

Thoughts?

yosifkit:
I think it may be worth considering.  I see why it takes so much longer; it forces a run of all the tests:
```console
Running code to generate profile data (this can take a while):
make run_profile_task
make[1]: Entering directory '/usr/src/python'
: # FIXME: can't run for a cross build
LD_LIBRARY_PATH=/usr/src/python ./python -m test.regrtest --pgo || true
Run tests sequentially
0:00:00 [  1/405] test_grammar
....
```

hannosch:
Yes, PGO means you need to compile all code once with profiling enabled, than execute some code that exercises the code in a typical way and finally compile everything again using the generated profile data as guidance. Python for now choose to use their regression test suite as the "typical code" to run and generate the profiling data.

Personally I don't think running all their tests is the best thing to do, but that's an issue to bring up with the Python folks and at least in my mind outside the scope of the docker image for it.

yosifkit:
It definitely is much longer:
```console
$ time docker build --no-cache 3.6-rc/
...
real	1m38.180s
user	0m0.157s
sys	0m0.069s

$ # after adding `--enable-optimizations`
$ time docker build 3.6-rc/
real	34m21.418s
user	0m0.267s
sys	0m0.124s

$ # the first build even includes the `apt-get update && apt-get install tcl tk` layer while the second used cache
```

fjorgemota:
Well,  we have another problem (besides the performance issue) to solve before enabling that flag.

[...]

PR #190, opened 2017-04-21
Fix segmentation fault when building Python 3.6 image

elbaschid:
I'm not 100% sure what the guidelines or rules are around PRs like this, happy to make any adjustments that might be required to get this merged.

This fixes the issues reported in #160 for the Alpine build. For Python
3.6, the build succeeds but test illustrate that `ctypes` related test
fail with segmentation faults. This also prevents optimized builds
because they rely on a full run of the test suite.

The fix is based on the release candidate build of Python 3.6.1 in
Alpine Edge. It switches to system libraries for `libffi` and `expat` in
the same way it's used in [the Alpine build
file](https://git.alpinelinux.org/cgit/aports/tree/main/python3/APKBUILD).

### DIFF
--- a/3.6/alpine/Dockerfile
+++ b/3.6/alpine/Dockerfile
@@ -58,11 +58,19 @@ RUN set -ex \
 		tk-dev \
 		xz-dev \
 		zlib-dev \
+		libffi-dev \
+		expat-dev \
 # add build deps before removing fetch deps in case there's overlap
 	&& apk del .fetch-deps \
 	\
 	&& cd /usr/src/python \
+	&& rm -r Modules/expat \
+		Modules/zlib \
+		Modules/_ctypes/darwin* \
+		Modules/_ctypes/libffi* \
 	&& ./configure \
+		--with-system-ffi \
+		--with-system-expat \
 		--enable-loadable-sqlite-extensions \
 		--enable-shared \
 	&& make -j$(getconf _NPROCESSORS_ONLN) \
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -52,11 +52,19 @@ RUN set -ex \
 		tk-dev \
 		xz-dev \
 		zlib-dev \
+		libffi-dev \
+		expat-dev \
 # add build deps before removing fetch deps in case there's overlap
 	&& apk del .fetch-deps \
 	\
 	&& cd /usr/src/python \
+	&& rm -r Modules/expat \
+		Modules/zlib \
+		Modules/_ctypes/darwin* \
+		Modules/_ctypes/libffi* \
 	&& ./configure \
+		--with-system-ffi \
+		--with-system-expat \
 		--enable-loadable-sqlite-extensions \
 		--enable-shared \
 	&& make -j$(getconf _NPROCESSORS_ONLN) \


tianon:
I'm glad you found a workaround for the segfault, but I'm still not sold on `--enable-optimizations` -- it's a _lot_ of additional build time with what seems like minimal benefit (as @yosifkit mentioned over in https://github.com/docker-library/python/issues/160).  It actually takes so long to build that Travis CI ends up killing the build. :disappointed:

elbaschid:
@tianon I can totally understand your point about the optimizations. The segmentation fault still remains as an issue though. If you are using the regular build an run the unittest, it basically fails with the same error as described in #160. 

How do you feel about removing `--enabled-optimizations` and `--with-lto` but merging the fix for the segmentation fault related to the ctypes build? I'm happy to update the PR accordingly.

3.6/alpine/Dockerfile
```diff
@@ -58,11 +58,19 @@ RUN set -ex \
...
+                libffi-dev \
+                expat-dev \
 # add build deps before removing fetch deps in case there's overlap
 	&& apk del .fetch-deps \
 	\
 	&& cd /usr/src/python \
+        && rm -r Modules/expat \
+		Modules/zlib \
+		Modules/_ctypes/darwin* \
+		Modules/_ctypes/libffi* \
```
yosifkit: There seems to be a mix of tabs and spaces throughout these changes; can you ensure to only use tabs. :pray: Thanks!

Dockerfile-alpine.template
```diff
@@ -52,11 +52,19 @@ RUN set -ex \
...
 	&& apk del .fetch-deps \
 	\
 	&& cd /usr/src/python \
+        && rm -r Modules/expat \
+		Modules/zlib \
+		Modules/_ctypes/darwin* \
+		Modules/_ctypes/libffi* \
 	&& ./configure \
+		--with-system-ffi \
+                --with-system-expat \
```
yosifkit: How will these changes affect the other versions of alpine python (3.3, 3.4, and 3.5)?

3.6/alpine/Dockerfile
elbaschid: Of course. Sorry about that. I've updated both files with tabs.

Dockerfile-alpine.template
elbaschid: The previous version of Python 3.x seemed to have worked without similar segmentation faults. But I've just run the regression tests in the Python 3.4 container with the settings above and they behaved the same way as without. So it does seem to work either way. I haven't tested all 3 versions, though. 

Do you want me to push updated Dockerfiles generated with the `update.sh` script and let them run on CI to test?

It would also be possible to use the docker `--build-arg` setting for running the unittest as part of the build process locally. For instance, adding this to the Alpine Dockerfiles:
```
...
ARG RUN_TESTS
...
	&& make -j$(getconf _NPROCESSORS_ONLN) \
        && if [ $RUN_TESTS ]; then LD_LIBRARY_PATH=/usr/src/python ./python -m test.regrtest; fi \
	&& make install \
...
```

And then running the docker command with `RUN_TESTS=1` would also run the tests without modifying any of the Docker images hosted on the Hub:

```
docker build --build-arg RUN_TESTS=1 --tag tested-python 3.4/alpine
```

Would that be something that makes sense to add for testing purposes? What do you suggest as the next steps?

PR #192, opened 2017-04-25
Add initial .appveyor.yml

tianon:
See also https://github.com/docker-library/openjdk/pull/106 and https://github.com/docker-library/golang/pull/156. :+1:


tianon:
(Travis is failing because I cancelled it -- nothing Travis or even Linux related is changed here.)

tianon:
GREEEN

PR #194, opened 2017-04-28
Switch to unconditional invocation of "get-pip.py"

tianon:
Combining `--without-ensurepip` / `Include_pip=0` and `get-pip.py`, we can move the `pip` installation to an entirely separate layer, thus allowing for `pip`-only bumps to build faster, given that Python will not also need to be recompiled.  This also insulates us from having to decide how often to bump `setuptools` (which has a fairly aggressive release cycle) -- it will naturally bump to the latest version at every `pip` release, every Python release, or every base-image-triggered rebuild.

See also https://github.com/docker-library/python/pull/186, especially https://github.com/docker-library/python/pull/186#issuecomment-297955361.

Thanks to @mal for bringing the issue up originally (and suggesting pretty much this exact fix in a slightly different form :innocent:), @JayH5 for pointing out how aggressive the `setuptools` release cycle is, and @dstufft for being generally awesome, doing good work, and on top of all that being willing to help point us in sane directions when we're going crazy. :heart: :heart:

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -16,11 +16,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ENV GPG_KEY C01E1CAD5EA2C4F0B8E3571504C367C218ADD4FF
 ENV PYTHON_VERSION 2.7.13
 
-# if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
-ENV PYTHON_PIP_VERSION 9.0.1
-ENV PYTHON_SETUPTOOLS_VERSION 35.0.2
-ENV PYTHON_WHEEL_VERSION 0.29.0
-
 RUN set -ex \
 	&& buildDeps=' \
 		tcl-dev \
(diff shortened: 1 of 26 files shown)


2.7/Dockerfile
```diff
@@ -42,30 +37,35 @@ RUN set -ex \
...
+	wget -O /tmp/get-pip.py 'https://bootstrap.pypa.io/get-pip.py'; \
+	\
+	python /tmp/get-pip.py "pip==$PYTHON_PIP_VERSION"; \
+	find /usr/local -depth \
+		\( \
+			\( -type d -a -name test -o -name tests \) \
+			-o \
+			\( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
+		\) -exec rm -rf '{}' +; \
+	rm -rf /tmp/get-pip.py ~/.cache
```
dstufft: You might just prefer adding ``--no-cache-dir`` to your ``get-pip.py`` call, that should tell pip not to use the cache at all.

tianon: Oooh, that sounds awesome :smile: -- does that mean that the "install python" layer might not need to purge `~/.cache` anymore either, now that it doesn't do anything pip-related?

dstufft: Probably? Not sure if anything else in the Python build process writes to ``~/.cache``.

tianon: Nice, it is the case :metal: (just tested) -- so the only file leftover is `~/.cache/pip/selfcheck.json`, which I got rid of by also adding `--disable-pip-version-check`! :balloon:

I'll update the PR shortly. :+1: :heart: :heart: :heart:

2.7/alpine/Dockerfile
```diff
@@ -89,6 +67,39 @@ RUN set -ex \
...
+# if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
+ENV PYTHON_PIP_VERSION 9.0.1
+
+RUN set -ex; \
+	\
+	apk add --no-cache --virtual .fetch-deps openssl; \
+	\
+	wget -O get-pip.py 'https://bootstrap.pypa.io/get-pip.py'; \
+	\
+	apk del .fetch-deps; \
```
JayH5: This spacing between the lines seems maybe a bit excessive..but nbd.

Also not sure about mixing the styles:
```dockerfile
RUN set -x \
  && foo \
...
```
with
```dockerfile
RUN set -ex; \
  foo; \
...
```
but that's really up to you.

tianon: Yeah, it's a bit odd, but it does help with the cross-variant diffs (doing `vimdiff` across all the different variants in this repo gets pretty tedious, so any little bit helps). :innocent: :+1:

Switching to semicolons is something I thought about doing to the whole file, but decided to keep the change smaller for now.  We've been trying to switch most things over to semicolons (slowly), especially after having fun with things like `foo && for xxx; do something-that-might-fail; done && bar`, which creates a "complex expression", and thus the rules for failure change (such that the `for xxx; do ...; done` bit of that expression will only fail if the final iteration fails, for example).

From https://www.gnu.org/software/bash/manual/bashref.html#index-set:

> The shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, part of any command executed in a && or || list except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !.

Thanks for taking a look :smile:

Issue #173, opened 2017-01-25
Update Alpine to 3.5


ghost:
https://hub.docker.com/_/alpine/

yosifkit:
See https://github.com/docker-library/python/pull/166, https://github.com/docker-library/python/pull/168, and https://github.com/docker-library/golang/pull/131#issuecomment-270677556.

lorddaedra:
:'-( 

May be add 
3.6.0-alpine, 3.6-alpine, 3-alpine, alpine (3.6/alpine/Dockerfile) / stay with 3.4 Alpine for now, add deprecation warning and switch to latest alpine later and after that always use latest alpine /
3.6.0-alpine3.4, 3.6-alpine3.4, 3-alpine3.4, alpine3.4 (3.6/alpine/Dockerfile)
3.6.0-alpine3.5, 3.6-alpine3.5, 3-alpine3.5, alpine3.5 (3.6/alpine/Dockerfile)

beaugunderson:
I would also use a `python:3.6-alpine3.5` tag if there was one (driven mostly by LibreSSL, postgres @ 9.6, and no warning during `update-ca-certs`) ✨

snimavat:
Plz update to alpine 3.5

nigma:
The `docker-library/golang` repository included the `-alpine3.5` variant (https://github.com/docker-library/golang/tree/master/1.7/alpine3.5) following the discussion in https://github.com/docker-library/golang/pull/131.

How about using the same approach here?

Lubo:
It was a big mistake not to go for Alpine 3.5 when Python 3.6 was released. Now, I'd prefer if we don't get too crazy with tags, but sticking to Alpine 3.4 for another year until Python 3.7 is released would suck big time.

tianon:
@Lubo Python 3.6 was released 2016-12-23; Alpine 3.5 was released 2016-12-22 -- we didn't even know we'd have to be reverting anything at that point, but yeah, we should've left 3.6 (a bit late now)

@nigma yeah, that's sane -- once https://github.com/docker-library/python/pull/194 settles, I'll take a look at adding `alpine3.5` variants here

Suor:
Trying to use Python 2.7 with R, which only available in alpine 3.5. Any reasons not to upgrade?

Lubo:
Alpine 3.6 is out a it's the new stable. Shouldn't we jump right there? Also, this issue has been around for nearly half a year now. What can we do speed the things up?

[...]

PR #166, opened 2016-12-28
Upgrade alpine to v3.5

PeterDaveHello:

PR #201, opened 2017-05-31
Add alpine 3.6 variant

skwashd:


tianon:
> See https://github.com/docker-library/python/pull/166, https://github.com/docker-library/python/pull/168, and https://github.com/docker-library/golang/pull/131#issuecomment-270677556.

skwashd:
Changes made as suggested by @tianon on [issue #173](https://github.com/docker-library/python/issues/173#issuecomment-305331880).

The [3.6-alpine3.6 build failed](https://travis-ci.org/docker-library/python/jobs/2394373340) while trying to verify the signature. This appears to be a transient issue. When I tried locally the first time it failed, 5 minutes later it worked. Last week [it built fine using alpine:3.6](https://travis-ci.org/docker-library/python/jobs/237793484) for all versions. The [2.7-alpine3.6 build](https://travis-ci.org/docker-library/python/jobs/239437357) passed. Can `$someone` please rerun the [239.3 build](https://travis-ci.org/docker-library/python/jobs/239437334)?

tianon:
Newer `gpg` is less forgiving about deleting the directory (socket files that we have a race condition with some agent cleaning up after itself while we delete) -- merging anyhow (will clean that up after). :+1:

JayH5:
Should this not be using LibreSSL instead of OpenSSL?

skwashd:
@JayH5 good point. I've submitted PR #203 to make that change. I'd appreciate any other feedback you have.

PR #203, opened 2017-06-06
Switch from OpenSSL to LibreSSL

skwashd:
This was suggested by @JayH5 on PR #201 after it was merged.

### DIFF
--- a/2.7/alpine3.6/Dockerfile
+++ b/2.7/alpine3.6/Dockerfile
@@ -17,7 +17,7 @@ ENV PYTHON_VERSION 2.7.13
 RUN set -ex \
 	&& apk add --no-cache --virtual .fetch-deps \
 		gnupg \
-		openssl \
+		libressl \
 		tar \
 		xz \
 	\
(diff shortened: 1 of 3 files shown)


JayH5:
That was fast. 😁  Thanks 🍰 

I'm not really sure what the LibreSSL vs OpenSSL policy is with Alpine official images. But it seems to me that if LibreSSL is the Alpine default then that is what we should use, especially if other packages are installed that depend on LibreSSL and that then conflicts with OpenSSL.

But I'm sure @tianon will have an answer for us as always :)

tianon:
In past conversions (when Alpine 3.5 switched from OpenSSL-by-default to LibreSSL-by-default) we only switched if packages we installed caused conflicts at build-time or run-time, but if this compiles and works, I'm +1 (especially since Alpine 3.5+ does use LibreSSL-by-default, so it makes sense for us to follow suit even/especially to ensure that compatibility works properly and packages can install).

tianon:
LGTM, in other words :+1:

PR #205, opened 2017-06-16
Fix _ctypes test failure on Alpine and build failure on Debian on ppc64le

yosifkit:
replaces/closes #190.

@elbaschid, sorry to replace your PR, I was working on multi-arch images and found out that this fixes the build for ppc64le (https://bugs.python.org/issue22157), so I applied it to all the versions of python 3.

This might fix the build failure on [arm64v8](https://doi-janky.infosiftr.net/job/multiarch/view/all-images/view/python/job/arm64v8/job/python/1/console), but I haven't tested it there yet.

### DIFF
--- a/3.3/Dockerfile
+++ b/3.3/Dockerfile
@@ -46,6 +46,8 @@ RUN set -ex \
 		--build="$gnuArch" \
 		--enable-loadable-sqlite-extensions \
 		--enable-shared \
+		--with-system-expat \
+		--with-system-ffi \
 		--without-ensurepip \
 	&& make -j "$(nproc)" \
 	&& make install \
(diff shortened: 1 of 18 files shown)


yosifkit:
Works for arm64v8 too! (python 3.3, testing others as well)

3.3/Dockerfile
```diff
@@ -46,6 +48,8 @@ RUN set -ex \
 		--build="$gnuArch" \
 		--enable-loadable-sqlite-extensions \
 		--enable-shared \
+		--with-system-ffi \
+		--with-system-expat \
```
tianon: `e` comes before `f`, doesn't it? :innocent:

yosifkit: I apparently sorted everything else but that :disappointed:

yosifkit:
All green! `libexpat1`, `libffi6`, their `-dev`s are already installed in the buildpack-deps image, so we can just use them in the non-slim Debian images. :tada:

Issue #104, opened 2016-04-24
Would a Stretch flavor be possible?


jrabbit:
I was using python:2.7-slim with stretch for a while and got hit by a version incompatibility b/t gnupg and taskd without a non-aptitude type solution.

yosifkit:
I think that right now Debian Stretch is too early in the cycle to promote as a stable base for language stacks.  Since you are already using `slim`, have you tried the `python:2.7-alpine` version?  It usually has newer packages to install (though it may require using the `edge` apk repository).

gcbirzan:
This should probably be revised, now that Stretch is frozen.

SpoonMeiser:
I would like to see a 2.7 stretch image too; we can't use the alpine images because they are incompatible with manylinux1 wheels.

daveisfera:
The list is of options is starting to get long, but what are the thoughts on having a slim variant based on stretch?

yosifkit:
We do have stretch based images for 2.7, 3.6, and 3.7 (https://github.com/docker-library/python/pull/215).  As for `slim` variant of stretch, we'll track that in #219.

PR #208, opened 2017-06-21
Add Debian Stretch variants

JayH5:
This adds a `stretch` variant to the `2.7`, `3.6`, and `3.6-rc` versions. These images are based on the `buildpack-deps:stretch` image.

I would _like_ the default Debian version to be Stretch, but that will probably break too much stuff. I'd also _like_ a slim Stretch image but that would probably create too many variations?

I copied the command to optionally install `gnupg2` and `dirmngr` from the GCC 4.9 image.


owais:
Having stretch as a variant would be great. jessie does not contain all new packages now and not everything is back ported. This doesn't have to become default. Just a variant would be great for now.

tianon:
Do you think we'd be able to get away without `python:2.7-stretch` for now? (probably not? :disappointed:)

We already support a fairly large number of variations here, and I'd like to try and keep that as limited as we can. :sweat_smile:

```console
$ bashbrew list --uniq python
python:2.7.13
python:2.7.13-slim
python:2.7.13-alpine
python:2.7.13-alpine3.6
python:2.7.13-wheezy
python:2.7.13-onbuild
python:2.7.13-windowsservercore
python:3.3.6
python:3.3.6-slim
python:3.3.6-alpine
python:3.3.6-wheezy
python:3.3.6-onbuild
python:3.4.6
python:3.4.6-slim
python:3.4.6-alpine
python:3.4.6-wheezy
python:3.4.6-onbuild
python:3.5.3
python:3.5.3-slim
python:3.5.3-alpine
python:3.5.3-onbuild
python:3.5.3-windowsservercore
python:3.6.1
python:3.6.1-slim
python:3.6.1-alpine
python:3.6.1-alpine3.6
python:3.6.1-onbuild
python:3.6.1-windowsservercore
python:3.6.2rc1
python:3.6.2rc1-slim
python:3.6.2rc1-alpine
python:3.6.2rc1-alpine3.6
python:3.6.2rc1-onbuild
python:3.6.2rc1-windowsservercore
```

JayH5:
It's tricky because different people have different needs... I don't particularly _need_ Stretch right now, I just wanted to get ahead of the curve. We do still maintain and run a lot of Python 2.7 software. We don't use the Wheezy images (and haven't ever) so I'd be fine with removing those variants. But that's probably not the case for some people 😞.

owais:
I personally do need stretch because it has never packages and libraries that are not backported. For example gdal which is a dependency for most GIS libraries and framworks like DjangoGIS. 

https://packages.debian.org/search?keywords=libgdal-dev. Jessie ships `1.10.1` while a stretch ships `2.1.2`. Right now, I must compile the newer version from source on Jessie based image. I'm sure there are more people who would like to have newer libs for their projects and also people who wouldn't want to get newer version without some advance warning so I think perhaps it's best to release another stretch based image and announce a date when default would become stretch and perhaps wheezy would be dropped.

Just my 2 cents.

antinescience:
I also need stretch; I've moved to alpine for a few services simply because Jessie packages are often well out of date at this point. That said, I think if you were looking to cut down on maintenance burden, you could reasonably say that variants for _new_ OS bases only support 2.7 and the most recent minor of 3 (so in this case, you'd have `python:2.7-stretch` and `python:3.6-stretch`). Supporting the last version of 2 in extended maintenance feels more important than supporting all the minors of 3.

tianon:
Doh, forgot about https://github.com/docker-library/python/issues/104 (folks asking for Stretch + 2.7 specifically).  I'm going to go remove those revert commits, rebase this, and get it merged.  As usual, thanks for your patience and your contribution, @JayH5! :heart:

tianon:
Oh, you did end up still including 2.7 -- I misread the revert commits. :+1:

I guess I'll just review this as-is instead! :sweat_smile:

PR #213, opened 2017-07-17
Correct operator precedence in find conditionals

### LINKED ISSUE #212: Bug in find conditional for tests/pyo/pyc deletion
The find command here aims to delete unwanted files:
https://github.com/docker-library/python/blob/761d766baa0ff33754436c0a1fd9aedc68cf7947/Dockerfile-debian.template#L52-L57

However the inner conditionals used are of form:
`-type d -a -name test -o -name tests`
and
`-type f -a -name '*.pyc' -o -name '*.pyo'`

And the [find man page](http://man7.org/linux/man-pages/man1/find.1.html) says:
> Please note that `-a` when specified implicitly (for example by two tests appearing without an
> explicit operator between them) or explicitly has higher precedence than `-o`.  This means that 
> `find . -name afile -o -name bfile -print` will never print `afile`.

...therefore they are interpreted like:
`(-type d -a -name test) -o -name tests`
and
`(-type f -a -name '*.pyc') -o -name '*.pyo'`

...which is presumably not what was intended.

I'll open a PR.

edmorley:
Previously the `type` option was only being applied to the first `-name` option, rather than both.

Fixes #212.

### DIFF
--- a/2.7/Dockerfile
+++ b/2.7/Dockerfile
@@ -48,9 +48,9 @@ RUN set -ex \
 	\
 	&& find /usr/local -depth \
 		\( \
-			\( -type d -a -name test -o -name tests \) \
+			\( -type d -a \( -name test -o -name tests \) \) \
 			-o \
-			\( -type f -a -name '*.pyc' -o -name '*.pyo' \) \
+			\( -type f -a \( -name '*.pyc' -o -name '*.pyo' \) \) \
 		\) -exec rm -rf '{}' + \
 	&& rm -rf /usr/src/python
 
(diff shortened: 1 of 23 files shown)


yosifkit:
Thanks!  Can you do a rebase? (3.6-rc was deleted)

tianon:
Took the liberty of rebasing it for you. :+1:

(Merging now without CI, because I didn't change any of the files and it was green before. :+1:)

edmorley:
Many thanks (got sidetracked hehe) :-)

PR #217, opened 2017-07-24
Add Debian Stretch variants for Python 3.4, 3.5

yggi49:
As there are now Debian Stretch variants for Python 2.7 and 3.6, this PR adds the corresponding variants for Python 3.4 and 3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,11 @@ env:
   - VERSION=3.6 VARIANT=jessie/slim
   - VERSION=3.6 VARIANT=alpine3.6
   - VERSION=3.6 VARIANT=alpine3.4
+  - VERSION=3.5 VARIANT=stretch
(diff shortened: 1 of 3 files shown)


JayH5:
@yggi49 have you seen the discussion in the PR for adding `stretch` variants for 2.7/3.6 (#208)? There is a concern that adding `stretch` variants for all the Python versions will result in too many different variants.

Also, the 3.4 build failed with what looks like an OpenSSL error. It's possible 3.4 has issues building against OpenSSL 1.1.0 in Stretch.

yggi49:
@JayH5 Thank you for drawing my attention to that other ticket; if too many variants are a maintenance concern, then it’s probably best to just wait until Stretch is made the default base.

The Python 3.4 issue is quite annoying, though, but it probably cannot be helped.  Thank you for pointing it out.

PR #228, opened 2017-10-08
Use TLS 1.2 in Windows builds

StefanScherer:
I tried to build Python with the microsoft/windowsservercore-insider build and found out that installing PIP aborted with "The request was aborted: Could not create SSL/TLS secure channel." It seems that https://bootstrap.pypa.io enforces TLS 1.2 only.

PowerShell still does not use TLS 1.2 by default, so here's a fix for all Windows Dockerfiles to do so. 

```
Step 7/8 : RUN Write-Host ('Installing pip=={0} ...' -f $env:PYTHON_PIP_VERSION); 	(New-Object System.Net.WebClient).DownloadFile('https://bootstrap.pypa.io/get-pip.py', 'get-pip.py'); 	python get-pip.py 		--disable-pip-version-check 		--no-cache-dir 		('pip=={0}' -f $env:PYTHON_PIP_VERSION) 	; 	Remove-Item get-pip.py -Force; 		Write-Host 'Verifying pip install ...'; 	pip --version; 		Write-Host 'Complete.';
 ---> Running in f1d24a39e5bd
Installing pip==9.0.1 ...
Exception calling "DownloadFile" with "2" argument(s): "The request was 
aborted: Could not create SSL/TLS secure channel."
At line:1 char:101
+ ... P_VERSION); (New-Object System.Net.WebClient).DownloadFile('https://b ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordE 
   xception
    + FullyQualifiedErrorId : WebException
 
The command 'powershell -Command $ErrorActionPreference = 'Stop'; Write-Host ('Installing pip=={0} ...' -f $env:PYTHON_PIP_VERSION); 	(New-Object System.Net.WebClient).DownloadFile('https://bootstrap.pypa.io/get-pip.py', 'get-pip.py'); 	python get-pip.py 		--disable-pip-version-check 		--no-cache-dir 		('pip=={0}' -f $env:PYTHON_PIP_VERSION) 	; 	Remove-Item get-pip.py -Force; 		Write-Host 'Verifying pip install ...'; 	pip --version; 		Write-Host 'Complete.';' returned a non-zero code: 1
```


StefanScherer:
AppVeyor is green, previous build showed the same error https://ci.appveyor.com/project/docker-library/python/build/build-74.master

tianon:
Arg, I was wondering why that build had started failing, but hadn't had a chance to look into it yet. :heart:

Would this also be fixable by switching to `Invoke-WebRequest` like we've switched to using elsewhere? (instead of instrumenting the .NET bits directly)

(ala https://github.com/docker-library/golang/blob/1116b4262228428be20d7e9413ad277c716adb16/Dockerfile-windows-nanoserver.template#L23)

StefanScherer:
No, even after changing to `Invoke-WebRequest` I had to add the security protocol to switch to TLS 1.2. This is really a quite weird default in current PowerShell.

tianon:
That's kind of bizarre -- is there an appropriate place for us to file an issue about it?

StefanScherer:
Probably https://github.com/PowerShell/PowerShell is a good place.
I'll update to Invoke-WebRequest, looks better :-)

Issue #236, opened 2017-10-25
stretch-slim: gnupg not removed when gnupg2 purged


JayH5:
The new `stretch-slim` images include a line like this in the Dockerfile to determine whether or not to install `gnupg2` and `dirmngr`:
```
$(command -v gpg > /dev/null || echo 'gnupg2 dirmngr')
```

These packages are later purged, but because of apt (?), the dependencies of `gnupg2` are not removed. To demonstrate:
```
❯ docker run --rm -it debian:stretch-slim bash
root@45ffcfc8aad0:/# apt-get update
...
root@45ffcfc8aad0:/# apt-get install -y --no-install-recommends gnupg2
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  gnupg gnupg-agent libassuan0 libksba8 libnpth0 libreadline7 libsqlite3-0 pinentry-curses readline-common
Suggested packages:
  parcimonie xloadimage dbus-user-session libpam-systemd pinentry-gnome3 scdaemon pinentry-doc readline-doc
Recommended packages:
  dirmngr gnupg-l10n
The following NEW packages will be installed:
  gnupg gnupg-agent gnupg2 libassuan0 libksba8 libnpth0 libreadline7 libsqlite3-0 pinentry-curses readline-common
0 upgraded, 10 newly installed, 0 to remove and 0 not upgraded.
Need to get 2977 kB of archives.
After this operation, 6081 kB of additional disk space will be used.
...
root@45ffcfc8aad0:/# apt-get purge -y --auto-remove gnupg2
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following packages will be REMOVED:
  gnupg2*
0 upgraded, 0 newly installed, 1 to remove and 0 not upgraded.
After this operation, 316 kB disk space will be freed.
(Reading database ... 6472 files and directories currently installed.)
Removing gnupg2 (2.1.18-8~deb9u1) ...
```

This has a few effects:
 1. `gnupg` and its dependencies remain installed taking up space.
 2. This `gnupg` doesn't work in a lot of cases because `dirmngr` will have been removed.
 3. The `command -v gpg` test done originally doesn't help to determine whether one should install `gnupg2` and `dirmngr` anymore, since the command is still available.

I'm not sure whether the solution is to install `gnupg` instead of `gnupg2` or to explicitly install both packages so that both are later removed.

JayH5:
[...]

PR #238, opened 2017-10-31
slim: Build against libncursesw5-dev

JayH5:
Fixes #235

Related docker-library/buildpack-deps#69

PR #239, opened 2017-11-01
Swap from "gnupg2" to "gnupg"

tianon:
See https://github.com/docker-library/python/issues/236 for details.

### DIFF
--- a/2.7/jessie/slim/Dockerfile
+++ b/2.7/jessie/slim/Dockerfile
@@ -37,7 +37,7 @@ RUN set -ex \
 		xz-utils \
 		zlib1g-dev \
 # as of Stretch, "gpg" is no longer included by default
-		$(command -v gpg > /dev/null || echo 'gnupg2 dirmngr') \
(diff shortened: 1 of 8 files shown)


tianon:
Doh, `stretch/slim` variants are now failing with the following:
```
	'python-imports' [6/8]...Traceback (most recent call last):
  File "./container.py", line 2, in <module>
    import readline
ImportError: libreadline.so.7: cannot open shared object file: No such file or directory
failed
```

tianon:
Kicked off https://travis-ci.org/docker-library/python/builds/296015684 to see if it's introduced by this PR or if it happens in current `master` :disappointed:

tianon:
https://github.com/docker-library/python/pull/238 had a totally passing build, so it can't possibly be the fault of that :disappointed: :scream: :confused:

tianon:
Oh!  `gnupg` has `dep: libreadline7 (>= 6.0)`, which we were unintentionally keeping, and it's now getting removed properly!  :metal: 

Will update in a moment.

PR #248, opened 2017-12-18
Alpine: Ensure thread stack size is large enough

### LINKED ISSUE #211: SEGFAULT when certain modules are used in a project with Alpine 3.6/Python 3.6
We've recently upgraded several of our containers to Python 3.6.

Some of the containers always segfault immediately after starting Django's `runserver` command.

In a couple of the containers upgrading our dependencies fixed the issue, but on some of the other containers that didn't have the issue upgrading the dependencies caused it to start happening.

In one example upgrading `requests` from 2.12.1 to 2.18.1 stopped the SEGFAULTs but adding the latest version of `boto3` caused them to start again.

The issue only happens when run under docker.

I have a [simple example repo here](https://github.com/beaugunderson/crash-test) and would like to offer a bounty of $100 to whoever can either fix the issue or explain it in a way that allows me to fix it. ✨

Edit: we've tested with `3-alpine` and `3-alpine3.6`; same result.

Also SEGFAULTs on `3.6.2rc2-alpine3.6`.

More clues: `requests` is also implicated in `boto3` which includes `botocore`--which includes a vendored `requests` 2.7.0.

JayH5:
Fixes #211 

Ports alpinelinux/aports@2026e12 to these images.

### DIFF
--- a/2.7/alpine3.4/Dockerfile
+++ b/2.7/alpine3.4/Dockerfile
@@ -60,6 +60,9 @@ RUN set -ex \
 		--enable-shared \
 		--enable-unicode=ucs4 \
 	&& make -j "$(nproc)" \
+# set thread stack size to 1MB so we don't segfault before we hit sys.getrecursionlimit()
+# https://github.com/alpinelinux/aports/commit/2026e1259422d4e0cf92391ca2d3844356c649d0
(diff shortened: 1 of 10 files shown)


JayH5:
docker-library/official-images#3813 should be merged before this so it can be tested.

tianon:
LGTM :+1:

PR #250, opened 2017-12-19
Bump python to 3.6.4

andriisoldatenko:
https://docs.python.org/3.6/whatsnew/changelog.html#python-3-6-4-final


tianon:
Updated in https://github.com/docker-library/python/commit/a1aa406bfd8c7b129e6e0ee0ba972b863624ac0d; thanks!

PR #251, opened 2017-12-19
Copy `python-gdb.py` to /usr/lib/debug to enable use with gdb

### LINKED ISSUE #89: include python-gdb.py in base containers
This file is relatively small, but helps a lot while debugging with gdb.  Even if it is compressed, it would help a lot!

It is in `/usr/src/python` before it gets wiped.  Will you accept a PR if I submit one?

https://docs.python.org/devguide/gdb.html#gdb-7-and-later

pbabics:
This is PR for https://github.com/docker-library/python/issues/89

When you want to use GDB with python docker images you have to create your own, because official does not have/support `python-gdb` script, it is removed during build time of image.

This PR proproses to copy it to `/usr/lib/debug/...`  which is in `auto-load` safe path. This enables to use commands like `py-bt` and others in GDB.

Note that to use GDB with `attach` you have to add `SYS_PTRACE` capability to container and even compile your own GDB against python source / library.

Closes #89

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -86,6 +86,7 @@ RUN set -ex \
 			-o \
 			\( -type f -a \( -name '*.pyc' -o -name '*.pyo' \) \) \
 		\) -exec rm -rf '{}' + \
+	&& cp /usr/src/python/python-gdb.py /usr/lib/debug/usr/local/bin/python$(echo "${PYTHON_VERSION}" | cut -d'.' -f1,2)-gdb.py \
 	&& rm -rf /usr/src/python
 
 # make some useful symlinks that are expected to exist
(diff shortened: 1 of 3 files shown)


tianon:
Sorry for the delay.

Are there other paths besides `/usr/lib/debug` that this file could be installed to?

When you say "and even compile your own GDB" -- is that required for this to work?  Does Debian's GDB not do the right thing?

pbabics:
There are more paths besides `/usr/lib/debug` but it may vary depending on base system.

GDB may not work when installed from debian package, because it is compiled against python from debian packages. But compiling GDB is pretty simple, so I would leave this for people who need it.

tianon:
Given that this is very complicated for us to get right, I'm going to (finally :bow:) close this for now, but I'll leave #89 open so that if/when we find a reasonable solution, this can be revisited. :heart:

PR #257, opened 2018-01-29
Update Dockerfile

JermineHu:
The code `ONBUILD COPY . /usr/src/app `  has been included requirements.txt file.


yosifkit:
Onbuild variants have been deprecated in https://github.com/docker-library/docs/pull/911. Besides, the order of the `COPY`, `RUN`, `COPY` is on purpose, so that when code is changed that doesn't affect `requirements.txt` the user doesn't have to wait for the pip install again when rebuilding the image.

Related issue https://github.com/docker-library/python/pull/196.

PR #259, opened 2018-02-08
Remove Windows variants of 3.5

tianon:
From https://www.python.org/downloads/release/python-355/:

> Also, Python 3.5.5 has only been released in source code form; no more official binary installers will be produced.


tianon:
See also https://www.python.org/ftp/python/3.5.4/ vs https://www.python.org/ftp/python/3.5.5/.

PR #266, opened 2018-03-01
Alpine: Add libs for nis module (Alpine 3.7/Python 3.7 only)

### LINKED ISSUE #247: NIS module included in Debian image but not in Alpine one
Hello, the installed Python on Alpine and Debian is not strictly the same.
I'm trying to use the `nis` module on Alpine, but it is not compiled.

```
$ docker run python:jessie python -c 'import nis; print("OK")'
OK
$ docker run python:stretch python -c 'import nis; print("OK")'
OK
$ docker run python:slim python -c 'import nis; print("OK")'
OK
$ docker run python:alpine python -c 'import nis; print("OK")'
ModuleNotFoundError: No module named 'nis'
```

I read somewhere that the Alpine libc does not include `nis` built-in. But `libnsl` is available on Alpine repos (https://pkgs.alpinelinux.org/package/v3.7/community/x86_64/libnsl).
Can we imagine to have this feature in the future 3.7 Alpine image?

Thanks,
Antoine

JayH5:
~~Bump Python 3.7 to 3.7.0b2~~

This kind of fixes #247, at least as best as is reasonably possible before Python 3.6.5/2.7.15.

The `nis` module is pretty obscure and I don't personally have a use for it, but it is the one and only optional module that wasn't building on Alpine. This adds ~2MB to the uncompressed image size:
```
REPOSITORY          TAG                        IMAGE ID            CREATED             SIZE
python              3.7.0b2-alpine3.7-no-nis   4ecaf70911d4        11 minutes ago      91.3MB
python              3.7.0b2-alpine3.7          902c6083404d        11 hours ago        93.6MB
```

### DIFF
--- a/3.7-rc/alpine3.7/Dockerfile
+++ b/3.7-rc/alpine3.7/Dockerfile
@@ -46,6 +46,8 @@ RUN set -ex \
 		gdbm-dev \
 		libc-dev \
 		libffi-dev \
+		libnsl-dev \
+		libtirpc-dev \
 		linux-headers \
 		make \
 		ncurses-dev \
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -40,6 +40,8 @@ RUN set -ex \
 		gdbm-dev \
 		libc-dev \
 		libffi-dev \
+		libnsl-dev \
+		libtirpc-dev \
 		linux-headers \
 		make \
 		ncurses-dev \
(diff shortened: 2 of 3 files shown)


JayH5:
@tianon ping on this if you have a second?

tianon:
Also relevant: https://github.com/tiran/peps/blob/ba2635cb59cd97264905ce32ef85f9456c1a5297/pep-9999.rst#nis (movement in Python upstream to remove `nis` entirely)

PR #274, opened 2018-03-29
3.6/alpine3.7: Build nis module now that 3.6.5 is patched

JayH5:
Continuation of #266.

Note also there's a Python 3.7.0b3 out now...~~guess the bot hasn't got to it yet.~~

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -160,8 +160,7 @@ for version in "${versions[@]}"; do
 
 		# Libraries to build the nis module available in Alpine 3.7, but also require this patch:
 		# https://bugs.python.org/issue32521
-		# TODO: Remove Python version check once 2.7 and 3.6 have the patch
-		if [[ "$variant" == alpine* ]] && [[ "$variant" != alpine3.7 || "$version" != 3.7* ]]; then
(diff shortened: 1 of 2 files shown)


update.sh
```diff
@@ -160,8 +160,7 @@ for version in "${versions[@]}"; do
 
 		# Libraries to build the nis module available in Alpine 3.7, but also require this patch:
 		# https://bugs.python.org/issue32521
-		# TODO: Remove Python version check once 2.7 and 3.6 have the patch
-		if [[ "$variant" == alpine* ]] && [[ "$variant" != alpine3.7 || "$version" != 3.7* ]]; then
+		if [[ "$variant" == alpine* ]] && [[ "$variant" != alpine3.7 ]]; then
```
yosifkit: Won't this also leave the libraries on `python:2.7-alpine3.7`?  Is that version fixed too?

JayH5: I did think so when I wrote this the first time, but the Python 2.7 Dockerfiles actually aren’t templated.

yosifkit: You are correct; I forgot about these lines https://github.com/docker-library/python/blob/33b16b4747706dd47ade94bebfde6b1144e4b758/update.sh#L135-L136

PR #276, opened 2018-04-10
Added support for nanoserver based windows containers

yorek:
Nanoserver has a much smaller footprint than window server core which make it a great option for any IoT scenario but also perfect anytime a full Windows Server core image is really not needed


yosifkit:
The official images do not yet support multistage builds. See https://github.com/docker-library/official-images/issues/3383.  Thanks, the Dockerfiles can still be useful for users that want to build their own `nanoserver` based images.

PR #278, opened 2018-04-25
Set PYTHONIOENCODING for Python 2

### LINKED ISSUE #147: Python 2.7 inside Docker cannot print unicode without --tty
I'm trying to run a simple Python app in Docker, using the official `python:2.7` image. 

My problem is that unless I enable `-t` / `--tty` in `docker run`, the python `print` command terminates when trying to print Unicode characters:

```
UnicodeEncodeError: 'ascii' codec can't encode characters...
```

If I enable `--tty` it works without errors.

The problem is that I want to run this process in the background, using `--detach` and that mode does not support `--tty`.

What kind of changes do I need to make to my Dockerfile so that it doesn't terminate when trying to print Unicode strings?

Minimal Dockerfile:

```
FROM python:2.7

CMD [ "python", "-c", "print (u'\\xc1')" ]
```

Minimal broken run command:

```
# error
docker run myimage 

# working
docker run --tty myimage
```

Note: it also works without problems in the official `python:3.5` image.

tianon:
Closes #147

### DIFF
--- a/2.7/alpine3.4/Dockerfile
+++ b/2.7/alpine3.4/Dockerfile
@@ -6,6 +6,8 @@ ENV PATH /usr/local/bin:$PATH
 # http://bugs.python.org/issue19846
 # > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
 ENV LANG C.UTF-8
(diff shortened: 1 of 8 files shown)


yosifkit:
Windows failure unrelated. LGTM

PR #280, opened 2018-04-27
adding GPG key servers

estevesd:
Sometimes when the default key server fails the built fails too.
This "fix" is inspired by what is done in [nginx official Dockerfile](https://github.com/nginxinc/docker-nginx)


estevesd:
I'm not able to run update.sh on my mac, and I don't have access to linux right now.
Therefore some tests are failing on the Dockerfiles which are not updated with the new template

PR #281, opened 2018-04-30
2.7/alpine3.7: Add nis libraries now that 2.7.15 is patched

JayH5:
The final fix for #247.

### DIFF
--- a/3.7-rc/alpine3.7/Dockerfile
+++ b/3.7-rc/alpine3.7/Dockerfile
@@ -47,12 +47,12 @@ RUN set -ex \
 		libc-dev \
 		libffi-dev \
 		libnsl-dev \
+		openssl \
+		openssl-dev \
 		libtirpc-dev \
 		linux-headers \
 		make \
 		ncurses-dev \
-		openssl \
-		openssl-dev \
 		pax-utils \
 		readline-dev \
 		sqlite-dev \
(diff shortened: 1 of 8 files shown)


3.7-rc/alpine3.7/Dockerfile
```diff
@@ -47,12 +47,12 @@ RUN set -ex \
 		libc-dev \
 		libffi-dev \
 		libnsl-dev \
+		openssl \
+		openssl-dev \
```
JayH5: I realize this now makes the openssl libs "unsorted", but the openssl builds should go away soon. Alpine 3.4 is being deprecated and Alpine 3.8's libressl should be supported by Python 3.7.

yosifkit:
Size difference for reference
```
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
python              2.7-alpine-test        bd330d7bf7a9        28 minutes ago      74.5MB
python              2.7-alpine3.7          ced9a5c17e81        4 weeks ago         69.8MB
```

tianon:
Also relevant: https://github.com/tiran/peps/blob/ba2635cb59cd97264905ce32ef85f9456c1a5297/pep-9999.rst#nis (movement in Python upstream to remove `nis` entirely)

PR #282, opened 2018-05-03
Add netbase so that `/etc/{protocols,rpc,services}` exists in slim images

### LINKED ISSUE #270: 3.x-slim are missing /etc/services which breaks getservbyname
This breaks `socket.getservbyname()` on those images, while that function works fine on `2.7-slim`.

```
docker run -ti python:3.6-slim ls -l /etc/services
ls: cannot access /etc/services: No such file or directory
```
and
```
docker run -ti python:3.5-slim ls -l /etc/services
ls: cannot access /etc/services: No such file or directory
```
vs
```
docker run -ti python:2.7-slim ls -l /etc/services
-rw-r--r-- 1 root root 19605 Oct 21  2014 /etc/services
```

yosifkit:
```bash
# Before (on 2.7 jessie slim):
After this operation, 5518 kB of additional disk space will be used.

# After:
After this operation, 5586 kB of additional disk space will be used.
```

Fixes #270. It is a minimal size change, but if it is not merged, it can be a good reference to users missing `/etc/protocols`, `/etc/rpc` or `/etc/services`.

Related discussion: https://github.com/debuerreotype/docker-debian-artifacts/issues/36#issuecomment-385212707

### DIFF
--- a/2.7/jessie/slim/Dockerfile
+++ b/2.7/jessie/slim/Dockerfile
@@ -16,6 +16,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		libreadline6 \
 		libsqlite3-0 \
 		libssl1.0.0 \
+		netbase \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GPG_KEY C01E1CAD5EA2C4F0B8E3571504C367C218ADD4FF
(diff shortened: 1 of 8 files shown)


yosifkit:
Whoops, apparently need this on non-slim, but my buildpack-deps images haven't rebuilt recently.  I am looking to add directly to buildpack-deps so that other languages don't run into the same issue so I think this one can be merged.

PR #283, opened 2018-05-03
Alpine: Don't install libraries for tkinter

JayH5:
* tkinter is rarely used in a container
* The Debian slim images also don't include these libraries (and also still build the module)
* Requires >12MB of libraries (x11, freetype...)
* The module is still built, so if needed, the required libraries can be installed and the module should still work

```
python        2.7-alpine3.7-no-tkinter   de65e3e2a0c5        About a minute ago   57MB
python        3.6-alpine3.7-no-tkinter   682fd4b35b37        12 minutes ago       74.8MB
python        2.7-alpine3.7              93c09abcfe57        6 days ago           69.5MB
python        3.6-alpine3.7              27e79c0fa4d2        12 days ago          87.4MB
```

### DIFF
--- a/2.7/alpine3.6/Dockerfile
+++ b/2.7/alpine3.6/Dockerfile
@@ -68,7 +68,10 @@ RUN set -ex \
 	&& make install \
 	\
 	&& runDeps="$( \
-		scanelf --needed --nobanner --format '%n#p' --recursive /usr/local \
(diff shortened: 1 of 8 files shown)


JayH5:
@tianon @yosifkit any chance one of you has a second to look at this and #281?

Thank you 🙏

PR #286, opened 2018-05-14
Add Alpine 3.6 and 3.7 variants for 3.5

matrixik:
OpenStack is now still stuck with Python 3.5 and we (Monasca) would like to use it with newer Alpine versions (especially that Alpine 3.4 got into EOS now).


matrixik:
Travis fail because of `gpg --keyserver ha.pool.sks-keyservers.net` problems, but when I run thsi commands locally they are fine.

matrixik:
Tests are flaky. This time none of my changes failed but only other tests.
There are problems with `ha.pool.sks-keyservers.net`.

matrixik:
Closing as it was fixed in #290

PR #290, opened 2018-05-22
Remove Alpine 3.4 variants (EOL)

### LINKED ISSUE #265: Tags `alpine` and `*-alpine` should refer to latest alpine release 3.7, not 3.4
I am requesting that the alpine related tags without explicit version always refer to the latest stable alpine release. Currently this should be 3.7.

In https://github.com/docker-library/python/blob/master/generate-stackbrew-library.sh#L17 this already seems to be done. But it's unclear to me why Python 3.6 is explicitly being mapped to Alpine 3.4:
https://github.com/docker-library/python/blob/master/generate-stackbrew-library.sh#L22

```
defaultAlpineVersion='3.7'
declare -A alpineVersions=(
	[2.7]='3.4'
	[3.4]='3.4'
	[3.5]='3.4'
	[3.6]='3.4'
```

Desired: `3-alpine` and `alpine` (and any other `-alpine`) tags should used latest stable python and alpine releases. Currently that is Python `3.6.4` and Alpine `3.7`.

Current: `3-alpine` and `alpine` tags use Alpine `3.4`, several versions behind.

tianon:
See https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases

Closes #265


JayH5:
Is it maybe worth waiting for Alpine 3.8 and using that for all the Python versions? Otherwise Python 3.4 & 3.5 users are left without an Alpine image. Alternatively, there is #286.

tianon:
This includes Alpine 3.7 on 3.4 and 3.5 (not entirely sold on the value of adding 3.6 at this point).

PR #303, opened 2018-07-06
Add Python 3.7/alpine3.8

### LINKED ISSUE #300: Python 3.7 Alpine 3.8
Since Alpine 3.8 has been released https://alpinelinux.org/posts/Alpine-3.8.0-released.html, it would be great to have python:3.7-alpine3.8

It contains some nice goodies, like newer OpenSSL.

Probably we would need to wait some days to see an official Alpine 3.8 image.

tao12345666333:
Add Alpine 3.8 and change openssl to libressl. Because Alpine 3.8 updated libressl to 2.7.4-r0 which Python 3.7's `ssl` module can compatible with it.

Now the official Alpine 3.8 is not not available.  https://github.com/gliderlabs/docker-alpine/issues/418

So I build one Alpine 3.8 image locally and test this change.

```
~ # cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.8.0
PRETTY_NAME="Alpine Linux v3.8"
HOME_URL="http://alpinelinux.org"
BUG_REPORT_URL="http://bugs.alpinelinux.org"
~ # python
Python 3.7.0 (default, Jul  6 2018, 06:55:30)
[GCC 6.4.0] on linux
>>> import ssl
>>> ssl.OPENSSL_VERSION
'LibreSSL 2.7.4'

```

After the Alpine 3.8 image be available, I will change `.travis.yml`

Fixes #300

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -112,7 +112,7 @@ for version in "${versions[@]}"; do
 	echo "$version: $fullVersion"
 
 	for v in \
-		alpine{3.6,3.7} \
+		alpine{3.6,3.7,3.8} \
 		{wheezy,jessie,stretch}{/slim,/onbuild,} \
 		windows/nanoserver-{1709,sac2016} \
 		windows/windowsservercore-{1709,ltsc2016} \
@@ -154,8 +154,8 @@ for version in "${versions[@]}"; do
 		esac
 
 		# https://bugs.python.org/issue32598 (Python 3.7.0b1+)
-		# TL;DR: Python 3.7+ uses OpenSSL functionality which LibreSSL doesn't implement (yet?)
-		if [[ "$version" == 3.7* ]] && [[ "$variant" == alpine* ]]; then
+		# TL;DR: Python 3.7+ uses OpenSSL functionality which LibreSSL 2.6.x in Alpine 3.7 doesn't implement
+		if [[ "$version" == 3.7* ]] && [[ "$variant" == alpine3.7 ]]; then
 			sed -ri -e 's/libressl/openssl/g' "$dir/Dockerfile"
 		fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ services: docker
 env:
   - VERSION=3.7 VARIANT=stretch
   - VERSION=3.7 VARIANT=stretch/slim
+  - VERSION=3.7 VARIANT=alpine3.8
   - VERSION=3.7 VARIANT=alpine3.7
   - VERSION=3.6 VARIANT=stretch
   - VERSION=3.6 VARIANT=stretch/slim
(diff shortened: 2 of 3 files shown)


tao12345666333:
The Alpine 3.8 image has been available. Updated `.travis.yml`.

JayH5:
@tao12345666333 

You'll need #304 and to update `update.sh` with something like:
```diff
diff --git a/update.sh b/update.sh
index 11683d6..7f53640 100755
--- a/update.sh
+++ b/update.sh
@@ -112,7 +112,7 @@ for version in "${versions[@]}"; do
 	echo "$version: $fullVersion"
 
 	for v in \
-		alpine{3.6,3.7} \
+		alpine{3.6,3.7,3.8} \
 		{wheezy,jessie,stretch}{/slim,/onbuild,} \
 		windows/nanoserver-{1709,sac2016} \
 		windows/windowsservercore-{1709,ltsc2016} \
@@ -154,8 +154,8 @@ for version in "${versions[@]}"; do
 		esac
 
 		# https://bugs.python.org/issue32598 (Python 3.7.0b1+)
-		# TL;DR: Python 3.7+ uses OpenSSL functionality which LibreSSL doesn't implement (yet?)
-		if [[ "$version" == 3.7* ]] && [[ "$variant" == alpine* ]]; then
+		# TL;DR: Python 3.7+ uses OpenSSL functionality which LibreSSL 2.6 in Alpine 3.7 doesn't implement
+		if [[ "$version" == 3.7* ]] && [[ "$variant" == alpine3.7 ]]; then
 			sed -ri -e 's/libressl/openssl/g' "$dir/Dockerfile"
 		fi
 
```

A Python 2.7/Alpine 3.8 image would be cool too 🙋🏼‍♂️

tao12345666333:
@JayH5  Thanks. I have updated it.

update.sh
```diff
@@ -154,8 +154,8 @@ for version in "${versions[@]}"; do
 		esac
 
 		# https://bugs.python.org/issue32598 (Python 3.7.0b1+)
-		# TL;DR: Python 3.7+ uses OpenSSL functionality which LibreSSL doesn't implement (yet?)
-		if [[ "$version" == 3.7* ]] && [[ "$variant" == alpine* ]]; then
+		# TL;DR: Python 3.7+ uses OpenSSL functionality which LibreSSL 2.6.x in Alpine 3.7 doesn't implement
+		if [[ "$version" == 3.7* ]] && [[ "$variant" == alpine3.7 ]]; then
```
lorddaedra: Should be `alpine3.8` instead of `alpine3.7`, isn't it?

tao12345666333: @lorddaedra  It isn't. On `alpine3.8` we can use `libressl` , the package's version is 2.7.x.

lorddaedra: looks like, you are right...

then we should merge this PR...

do we have any reasons to wait for anything?.. last commit was added 6 days ago...

lorddaedra: looks like, you are right...

then we should merge this PR...

do we have any reasons to wait for anything?..

pvizeli:
Can we have also python 3.6 with alpine 3.8 ?

tao12345666333:
@pvizeli  Yes, we can. I'll be doing one shortly.

tianon:
https://github.com/docker-library/python/pull/311 :+1:

PR #307, opened 2018-07-10
alpine: Don't install ca-certificates on Alpine 3.7+

JayH5:
* CA certs come bundled on Alpine 3.7+
* On Alpine 3.6, rather install `libressl` which also include CA certs, allows us to use wget, and doesn't increase the size over `ca-certificates`

No significant change to image size (decreases ~0.7MB). Mostly a cleanup.


JayH5:
@tianon I've rebased this on top of #311

JayH5:
Rebased on top of #315

tianon:
Thanks @JayH5 :heart:

(especially for your patience :sweat_smile:)

PR #313, opened 2018-07-20
Update defaultAlpineVersion, from v3.7 to v3.8

PeterDaveHello:
Alpine Linux v3.8 was released for weeks, reference to the time for changing `defaultAlpineVersion` from `3.6` to `3.7`, I guess it's time to bump the default Alpine based image to v3.8?

Thanks a lot!

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -8,7 +8,7 @@ declare -A aliases=(
 )
 
 defaultDebianSuite='stretch'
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
Thanks @tianon! Do you have schedule to update this to Docker Hub? Not sure if it's still the time to update https://github.com/docker-library/official-images/pull/4622/ for it, thanks for your effort on those projects anyway!

PR #315, opened 2018-07-20
Add templates for Python 2.7

tianon:
(This allows us to generate ~10 files from 4, which is a pretty decent ratio IMO -- it also helped me catch a few bits I missed previously.)

Hopefully now @yosifkit will stop complaining at me every time he updates Python (not to mention my own cursing). :smile: :innocent:

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -132,21 +132,12 @@ for version in "${versions[@]}"; do
 			# use "debian:*-slim" variants for "python:*-slim" variants
 			tag+='-slim'
 		fi
+		if [[ "$version" == 2.* ]]; then
+			template="caveman-${template}"
+		fi
 		template="Dockerfile-${template}.template"
 
-		if [[ "$version" == 2.* ]]; then
-			fail=1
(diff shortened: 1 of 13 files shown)


update.sh
```diff
@@ -132,21 +132,12 @@ for version in "${versions[@]}"; do
...
-				[[ "$newVersion" == 3.* ]] || continue
-				[ -f "$newVersion/$v/Dockerfile" ] || continue
-				echo "  TODO: vimdiff $newVersion/$v/Dockerfile $version/$v/Dockerfile"
-				fail=
-				break
-			done
-			[ -z "$fail" ] || { echo >&2 "error: cannot find a suitable version for 'vimdiff' on '$version/$v/Dockerfile'"; exit 1; }
-		else
-			{ generated_warning; cat "$template"; } > "$dir/Dockerfile"
+			template="caveman-${template}"
```
yosifkit: `caveman` :laughing: :rofl: :+1:

tianon: image

PR #316, opened 2018-07-23
Reduce clone depth of official-images on Travis CI

PeterDaveHello:
Only the latest version of official-images repo is needed. We can speed up this part in the CI build.
(As you can see that the repository already has almost 9k commits)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ matrix:
       env: VERSION=2.7 VARIANT=alpine3.9
 
 install:
-  - git clone https://github.com/docker-library/official-images.git ~/official-images
+  - git clone --depth 1 https://github.com/docker-library/official-images.git ~/official-images
   - if [ "$TRAVIS_OS_NAME" = 'linux' ]; then wget -qO- 'https://github.com/tianon/pgp-happy-eyeballs/raw/master/hack-my-builds.sh' | bash; fi
 
 before_script:


tianon:
Checking on the logs for the real-world time this line takes, the difference is ~1.21s before and ~1.09s after, so I'm not sure this is really necessary.

tianon:
For reference:

- Before: https://travis-ci.org/docker-library/python/jobs/407300961#L364
- After: https://travis-ci.org/docker-library/python/jobs/407008409#L372

tao12345666333:
@tianon  I think the ~0.12s for Travis CI build jobs is no effect. And for human this is not really necessary.

PeterDaveHello:
I believe that various. The variation on Travis CI is not quite small, if the network is busier, the difference could be more notable, though it's still no really big deal I agree.

jrbenito:
Hi,
We should think this not as simple matter of time but a matter of waste resource. Why download entire history if needs only the last commit? Besides the 0.12s there is bandwidth costs applied to Travis and Github links. Why not play easy on bandwidth whenever possible?

I agree with @PeterDaveHello  that time may vary according to network traffic and size of the repository.

tianon:
Ok, bandwidth comparison is ~6.45MB vs ~489.32KB (which is still pretty negligible either way).

Even at 6.45MB, that's ~158 builds before we even reach a gigabyte of bandwidth.

tianon:
To be clear: I'm not entirely opposed to this change, but I don't want to cascade this update to every other `.travis.yml` under https://github.com/docker-library because I don't see a compelling reason to generate so much churn.

tao12345666333:
@tianon  Yeah, I think that's too small a gain compared to the cost of modifying it.

PeterDaveHello:
We growth of the `official-images`, the effect will actually being a little more bigger, currently, it takes > 10MB per clone, but anyway, I'll respect decision from maintainer, so won't send this PR to other repos. @tianon I've rebased and resolved the conflicts here, is it good to go now? Thanks.

.travis.yml
```diff
@@ -68,7 +68,7 @@ matrix:
       env: VERSION=2.7 VARIANT=alpine3.9
 
 install:
-  - git clone https://github.com/docker-library/official-images.git ~/official-images
+  - git clone --depth 1https://github.com/docker-library/official-images.git ~/official-images
```
ulgens: Missing space?

PeterDaveHello: oh, yes, break in accident when resolving conflicts, thanks!

RDIL:
I think this should be good to go if conflicts are resolved :+1:

yosifkit:
We've moved on to using github actions(https://github.com/docker-library/python/pull/473), so this is not needed.

PR #317, opened 2018-07-25
Update Lisence

beaglebao:
Update the year of Lisence


tianon:
I'd prefer to go in the direction of https://github.com/docker-library/golang/pull/212 (removing explicit "end" years) -- would you like to update, or would you rather I take over from here?

tianon:
https://github.com/docker-library/python/commit/663b4a6290f69446ee4c77cbeac079fba4a708c6 :+1:

PR #326, opened 2018-08-03
Python 3.7: Install libuuid packages for native _uuid module

JayH5:
Still to do:
- [x] Get this working on the non-slim Debian images
- [x] Make the update.sh code cleaner?

I'm not sure how best to do the above?

The `util-linux-dev` package was actually already being installed (as a dependency of `libtirpc-dev`) on Alpine so the native module was already being built and you can see `libuuid` as one of the final Python runtime dependencies. This makes it explicit and gets Debian to build the module as well.

### DIFF
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -12,6 +12,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # Python 3.4 on Stretch+ needs to use an older version of "libssl-dev" (these lines both get removed for every other combination)
 		libssl-dev \
 		tk-dev \
+		uuid-dev \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GPG_KEY %%PLACEHOLDER%%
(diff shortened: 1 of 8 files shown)


JayH5:
On my machine, running benchmarks like [these](https://bugs.python.org/issue11063#msg303284), I got ~20% speedup on both UUID1 generation and module import. So not a huge increase on already quite fast operations.

Dockerfile-debian.template
```diff
@@ -12,6 +12,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # Python 3.4 on Stretch+ needs to use an older version of "libssl-dev" (these lines both get removed for every other combination)
 		libssl-dev \
 		tk-dev \
+		uuid-dev \
```
tianon: Hah, we finally found a popular `-dev` that isn't already part of `buildpack-deps`! :smile:

If we can find more examples of it being used (Ruby gems, etc), I'd be +1 to adding it to `buildpack-deps` itself, but that's a separate discussion. :innocent:

PR #330, opened 2018-08-14
Add --no-cache for all apk commands

nvllsvm:
The Alpine images contain unnecessary repository indexes in `/var/cache/apk`. The following changes were made to ensure they are not saved.

- ensure --no-cache to all `apk add` commands
- add --no-cache to all `apk del` commands. Even though it removes packages, running it will ensure that an index is present locally.

### DIFF
--- a/2.7/alpine3.6/Dockerfile
+++ b/2.7/alpine3.6/Dockerfile
@@ -79,7 +79,7 @@ RUN set -ex \
 		| tr ',' '\n' \
 		| sort -u \
 		| awk 'system("[ -e /usr/local/lib/" $1 " ]") == 0 { next } { print "so:" $1 }' \
-		| xargs -rt apk add --virtual .python-rundeps \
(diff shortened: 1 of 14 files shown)


tianon:
This is actually a (recent) bug, and they're planning to fix it: https://bugs.alpinelinux.org/issues/9063

(https://git.alpinelinux.org/cgit/apk-tools/commit/?id=e1a05c74cb8ae6ab41ee960343cc64a0a9af95a5)

https://git.alpinelinux.org/cgit/apk-tools/commit/?id=8a28c6d0d4f81406c67e76b2bcdbe491aa734717 appears to be the commit which introduced this "autoupdate" behavior.

nvllsvm:
@tianon - Updated PR to only address the missing for apk add commands.

yosifkit:
Added commit running `update.sh` so that travis will run with the changes.
While this only matters for alpine 3.8, there is no material change on older versions, so LGTM.

Issue #334, opened 2018-09-14
IPV6 Issues with ha.pool.sks-keyservers.net


austinpray:
Have multiple people getting:

```
+ gpg --keyserver ha.pool.sks-keyservers.net --recv-keys 0D96DF4D4110E5C43FBFB17F2D347EA6AA65421D
gpg: keybox '/tmp/tmp.7GlY5wPunO/pubring.kbx' created
gpg: keyserver receive failed: Cannot assign requested address
```

When trying to build this image

See: https://gitlab.com/meltano/meltano/merge_requests/53
See: https://github.com/nodejs/docker-node/issues/380

So we need to do something like https://gitlab.com/meltano/meltano/blob/443ea69880dc7a86f099df2ddc0a7718a7587d61/docker/vendor/python/keyservers.patch

tianon:
See https://github.com/docker-library/python/pull/311 for how we combat this (both in Travis and on the official build servers).

PR #335, opened 2018-09-14
Adds keyserver fallbacks

austinpray:
Closes https://github.com/docker-library/python/issues/334
Ref https://github.com/nodejs/docker-node/issues/380
Ref https://github.com/nodejs/docker-node/blob/526c6e618300bdda0da4b3159df682cae83e14aa/Dockerfile-alpine.template#L23-L25
Ref https://gitlab.com/meltano/meltano/issues/90


tianon:
See https://github.com/docker-library/official-images/issues/4252#issuecomment-405749749.

austinpray:
Ah I missed that. Thanks for the info. I'll adjust accordingly

PR #339, opened 2018-09-24
Remove unnecessary package definitions to install

eavelardev:
the tk-dev installation includes dpkg-dev, libc6-dev,  libexpat1-dev, make and zlib1g-dev packages so their definitions are unnecessary.

It's a suggested change and more slim :).


yosifkit:
We'd rather be explicit about the packages we need rather than relying on the dependencies of another package we also install (which could change).

PR #311, opened 2018-07-13
Refactor dependencies handling to keep necessary runtime dependencies automatically

tianon:
(accounting for "tkinter" and "libssl" edge cases appropriately)

Closes #309
Closes #306
Closes #303
Closes #283

(combined/refactored)

Closes #300

PR #348, opened 2018-10-30
GPG Keyserver to support Enterprise Firewalls

cmaenner:
Added support to allow `gpg --keyserver` to use port 80 to work through enterprise firewalls


tianon:
See https://github.com/docker-library/python/pull/311, https://github.com/docker-library/python/pull/280, https://github.com/docker-library/official-images/issues/4252 for how we combat this for all our own builds (both in Travis and downstream on the official-images build servers).

PR #358, opened 2018-12-04
Use system FFI and expat

blopker:
Just the Python 2.7 fixes from https://github.com/docker-library/python/pull/357

### DIFF
--- a/2.7/alpine3.6/Dockerfile
+++ b/2.7/alpine3.6/Dockerfile
@@ -44,10 +44,12 @@ RUN set -ex \
 		bzip2-dev \
 		coreutils \
 		dpkg-dev dpkg \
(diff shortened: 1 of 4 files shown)


yosifkit:
What does this fix?

Are the versions bundled with python too old to be useful? Are the versions in Alpine 3.6, 3.7, and 3.8 actually better?

blopker:
It's the same fix you made here: https://github.com/docker-library/python/pull/205 Just added to Python 2.7. The same issue came up when running the tests for PGO in https://github.com/docker-library/python/pull/357 during the build. Python 2.7 segfaulted while testing ctypes.

blopker:
The Debian versions didn't segfault, so I didn't change them.

PR #375, opened 2019-01-31
Alpine 3.9

JayH5:
Adds Alpine 3.9 variants for all versions of Python. Alpine 3.9 switches back to OpenSSL instead of LibreSSL.

My only concern with this is whether it's still worth supporting Python 3.4. pip 19 already gives a warning prompt when run under Python 3.4:
```
DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).
```
...and Python 3.4 doesn't support the version of OpenSSL in Alpine 3.9, so requires LibreSSL.


tao12345666333:
LGTM :+1:

jeanphix:
Can't wait for this one :+1:

ivlevdenis:
:rocket:

tianon:
Seems good, thanks! :+1:

I've pushed an extra commit to sort the package lists and another to remove older Alpine versions (since we're only committing to support 2 per Python version now), and I think your solution for the almost-EOL Python 3.4 is reasonable. :+1: :heart:

Hopefully Travis will agree with this so we can merge it. :+1:

JayH5:
Sorting the packages 🤦‍♂️ there's always something I miss 😅. Thanks!

PR #404, opened 2019-07-09
Add "--enable-optimizations" (and a slimmer PROFILE_TASK than the default of "run all tests")

### LINKED ISSUE #160: Consider adding '--enable-optimizations' configure argument

tianon:
Closes #160

Thanks to @gpshead for helping us figure out `PROFILE_TASK` (and for providing a suitable list for us to start with). :+1:

### DIFF
--- a/3.5/alpine3.10/Dockerfile
+++ b/3.5/alpine3.10/Dockerfile
@@ -67,6 +67,7 @@ RUN set -ex \
 	&& ./configure \
 		--build="$gnuArch" \
 		--enable-loadable-sqlite-extensions \
+		--enable-optimizations \
 		--enable-shared \
 		--with-system-expat \
 		--with-system-ffi \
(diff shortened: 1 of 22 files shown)


tianon:
Looking at Travis, it appears the average is ~2-3x, which is still acceptable IMO.

PR #408, opened 2019-07-23
Update PIP to 19.2.1

### LINKED ISSUE #407: Verify get-pip.py
Currently `get-pip.py` is downloaded from a remote server and installed without checking that it's legitimate [1]

I'm aware that there's no official way to verify the downloaded file and also that it might change at any time, so IMHO it might be good to keep a local, known to be good, version of that file and update it from time to time.

[1] https://github.com/docker-library/python/blob/13ae0c0b3bf09ad3cf2f75f5e9a4ab2415a4b109/3.7/stretch/Dockerfile#L110

tianon:
Also, use a fixed commit from https://github.com/pypa/get-pip and add a sha256 checksum for verification.

Closes #407

(This is likely going to have conflicts very shortly given that PIP is updated and our automated job is running, so I'll try to keep up with rebasing it.)

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -34,6 +34,9 @@ fi
 versions=( "${versions[@]%/}" )
 
 pipVersion="$(curl -fsSL 'https://pypi.org/pypi/pip/json' | jq -r .info.version)"
+getPipCommit="$(curl -fsSL 'https://github.com/pypa/get-pip/commits/master/get-pip.py.atom' | tac|tac | awk -F '[[:space:]]*[<>/]+' '$2 == "id" && $3 ~ /Commit/ { print $4; exit }')"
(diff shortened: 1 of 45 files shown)


update.sh
```diff
@@ -34,6 +34,9 @@ fi
 versions=( "${versions[@]%/}" )
 
 pipVersion="$(curl -fsSL 'https://pypi.org/pypi/pip/json' | jq -r .info.version)"
+getPipCommit="$(git ls-remote https://github.com/pypa/get-pip.git refs/heads/master | cut -d$'\t' -f1)"
```
yosifkit: Can this be restricted to just `get-pip.py` so that the commit doesn't get updated on an unrelated change in one of the folders?

tianon: Updated. :+1:

PR #411, opened 2019-08-01
Remove explicit PROFILE_TASK for 3.8+

tianon:
The upstream default has been slimmed down to a reasonable amount in Python 3.8+!

See also:

- https://bugs.python.org/issue36044
- https://github.com/docker-library/python/issues/160
- https://github.com/python/cpython/pull/14702
- https://github.com/python/cpython/pull/14910


tianon:
Travis before (https://travis-ci.org/docker-library/python/builds/566732303):

image


Travis after (https://travis-ci.org/docker-library/python/builds/566732303):

image

PR #423, opened 2019-10-14
Update to 3.8.0 (GA)

tianon:


hannseman:
When will it be available on Docker Hub?

tianon:
https://github.com/docker-library/faq#an-images-source-changed-in-git-now-what

PR #366, opened 2019-01-20
adds USDT trace points for supported distros

tgross:
Python supports user-level statically defined tracing (USDT) for platforms that support it. This has historically been not compiled-in by default in Linux distributions because only hosts with `systemtap` can take advantage of it. Newer Linux kernels  (including all kernels supported by Docker) have eBPF, and this allows for all users to take advantage of tools like [pythonflow](https://github.com/iovisor/bcc/blob/master/tools/pythonflow.sh), [pythongc](https://github.com/iovisor/bcc/blob/master/tools/pythongc.sh), etc. Including this option would have the happy side-effect of allowing for the USDT probes to be used for DTrace on platforms where Docker containers can be run on Illumos (ex. Joyent's Triton).

The upstream implementation is a little weird (ref https://docs.python.org/3/howto/instrumentation.html). The `--with-dtrace` flag expects the `dtrace` binary to exist so that appropriate header files can be written. The `systemtap-sdt-dev` package for Debian provides a mock "`dtrace` binary" (actually a Python script) which can generate the header files.

This package or one like it is not available for Alpine. I have [another branch](https://github.com/tgross/docker-python/pull/2) with an experiment to vendor the header files and the mock `dtrace` binary from the Debian `systemtap-sdt-dev` package. This works inasmuch as the header gets created and I've verified the USDT probes work, but it's a totally gross approach and perhaps dubious from a licensing standpoint so I didn't include it in this PR. But I'd love to have a discussion about what the preferred approach would be.

PR #433, opened 2019-11-15
KM-376 built a debug version of the python-3.6 docker imnage, on whic…

mfletcher-grove:
…h we rely, so that we can use debugging tools in docker containers. First commit, not yet proven. See https://blog.0x74696d.com/posts/debugging-python-containers-in-production/.


tianon:
See https://github.com/docker-library/python/pull/366 (of which I believe this is a duplicate). :disappointed:

Issue #444, opened 2019-12-21
Support for socket AF_BLUETOOTH


sergioisidoro:
Python 3 has documentation for support of Bluetooth with `AF_BLUETOOTH` socket -- [docs](https://docs.python.org/3/library/socket.html)

However, it seems that (at least for linux) it needs bluetooth.h (from e.g. libbluetooth-dev) for that to work, otherwise we get 

`AttributeError: module 'socket' has no attribute 'AF_BLUETOOTH'`

Some other issues mentioning this case are, for example 

* https://github.com/ContinuumIO/anaconda-issues/issues/1035
* https://stackoverflow.com/questions/29107537/missing-socket-af-bluetooth-in-anaconda-python

I was wondering if it is possible to add `bluetooth.h` to the build deps, since this part of the standard documentation

wglambert:
Closing with the merge of https://github.com/docker-library/python/pull/445

PR #445, opened 2019-12-21
Support for socket AF_BLUETOOTH in python standard lib

sergioisidoro:
### ⚠️ Disclaimer: I'm new around here. ⚠️

As mentioned in https://github.com/docker-library/python/issues/444 , python 3 has native support for bluetooth sockets, with [socket family AF_BLUETOOTH](https://docs.python.org/3/library/socket.html#socket-families)

However, it only has support for it if built with `bluetooth.h` in the dependencies 

Before:
```
└─▪docker run -it python:3.7 python 
Unable to find image 'python:3.7' locally
3.7: Pulling from library/python
16ea0e8c8879: Pull complete 
50024b0106d5: Pull complete 
ff95660c6937: Pull complete 
9c7d0e5c0bc2: Pull complete 
29c4fb388fdf: Pull complete 
8659dae93050: Pull complete 
c8208a94eb32: Pull complete 
465b26a31c40: Pull complete 
84af071ddf6f: Pull complete 
Digest: sha256:4ce3a6ba2c2103d50ac59d18d46a484ad1b489b89609f548ef78f02d43eaecf9
Status: Downloaded newer image for python:3.7
Python 3.7.6 (default, Dec 20 2019, 22:36:14) 
[GCC 8.3.0] on linux
>>> from socket import AF_BLUETOOTH"
  File "<stdin>", line 1
    from socket import AF_BLUETOOTH"
                                   ^
SyntaxError: EOL while scanning string literal
>>> from socket import AF_BLUETOOTH
ImportError: cannot import name 'AF_BLUETOOTH' from 'socket' (/usr/local/lib/python3.7/socket.py)
```

After:
```
└─▪docker run -it sergio/py37 python
Python 3.7.6 (default, Dec 21 2019, 23:33:58) 
[GCC 8.3.0] on linux
>>> from socket import AF_BLUETOOTH
>>> AF_BLUETOOTH
<AddressFamily.AF_BLUETOOTH: 31>
>>> 
```

This might be a bit too much to add to the `slim` images, but since this is a native python support, I was wondering if it would make sense to include in these images. 

Some mentions of the issue:
* https://github.com/ContinuumIO/anaconda-issues/issues/1035
* https://stackoverflow.com/questions/29107537/missing-socket-af-bluetooth-in-anaconda-python
* https://gitlab.alpinelinux.org/alpine/aports/issues/7352

~~I have not ran `update.sh`~~. I have ran update.sh but the pip version commit hash is broken. Had to hard-code it to generate the Dockerfiles

This change does not apply to 2.x, but since 2.x is deprecated in 9 days anyway

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -56,6 +56,7 @@ RUN set -ex \
 		util-linux-dev \
 		xz-dev \
 		zlib-dev \
+		bluez-dev \
 # add build deps before removing fetch deps in case there's overlap
 	&& apk del --no-network .fetch-deps \
 	\
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -32,6 +32,7 @@ RUN set -ex \
 	&& rm python.tar.xz \
 	\
 	&& apk add --no-cache --virtual .build-deps  \
+		bluez-dev \
 		bzip2-dev \
 		coreutils \
 		dpkg-dev dpkg \
@@ -56,7 +57,6 @@ RUN set -ex \
 		util-linux-dev \
 		xz-dev \
 		zlib-dev \
-		bluez-dev \
 # add build deps before removing fetch deps in case there's overlap
 	&& apk del --no-network .fetch-deps \
 	\
--- a/Dockerfile-caveman-slim.template
+++ b/Dockerfile-caveman-slim.template
@@ -13,6 +13,7 @@ ENV PYTHONIOENCODING UTF-8
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		ca-certificates \
 		netbase \
+		bluez-dev \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GPG_KEY %%PLACEHOLDER%%
--- a/Dockerfile-caveman-slim.template
+++ b/Dockerfile-caveman-slim.template
@@ -13,7 +13,6 @@ ENV PYTHONIOENCODING UTF-8
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		ca-certificates \
 		netbase \
-		libbluetooth-dev \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GPG_KEY %%PLACEHOLDER%%
--- a/Dockerfile-caveman-slim.template
+++ b/Dockerfile-caveman-slim.template
@@ -13,7 +13,7 @@ ENV PYTHONIOENCODING UTF-8
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		ca-certificates \
 		netbase \
-		bluez-dev \
+		libbluetooth-dev \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GPG_KEY %%PLACEHOLDER%%
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -13,7 +13,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		libssl-dev \
 		tk-dev \
 		uuid-dev \
-		bluez-dev \
+		libbluetooth-dev \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GPG_KEY %%PLACEHOLDER%%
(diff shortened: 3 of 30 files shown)


sergioisidoro:
Update: On Ubuntu 18 python comes built with bluetooth support, so this change makes the behaviour similar to, for example, doing a `apt install python`

Dockerfile-slim.template
```diff
@@ -38,6 +38,7 @@ RUN set -ex \
 		wget \
 		xz-utils \
 		zlib1g-dev \
+		libbluetooth-dev \
```
tianon: This list should be in sorted order. :+1:

Dockerfile-debian.template
```diff
@@ -13,6 +13,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		libssl-dev \
 		tk-dev \
 		uuid-dev \
+		libbluetooth-dev \
```
tianon: This list should be in sorted order. :+1:

Dockerfile-caveman-slim.template
```diff
@@ -13,6 +13,7 @@ ENV PYTHONIOENCODING UTF-8
 RUN apt-get update && apt-get install -y --no-install-recommends \
 		ca-certificates \
 		netbase \
+		libbluetooth-dev \
```
tianon: This should be down in the temporary list (not this persistent list), like in `Dockerfile-slim.template` -- also, is it confirmed that Python 2 does support this as well or is it Python 3 only?

Dockerfile-alpine.template
```diff
@@ -56,6 +56,7 @@ RUN set -ex \
 		util-linux-dev \
 		xz-dev \
 		zlib-dev \
+		bluez-dev \
```
tianon: This list should be in sorted order. :+1:

Dockerfile-caveman-slim.template
tianon: Doh, just saw your note ("This change does not apply to 2.x, but since 2.x is deprecated in 9 days anyway") -- so this should be removed from all `-caveman` templates which should remove it from all 2.x versions. :+1:

yosifkit:
Since https://github.com/docker-library/python/pull/442 was merged this now needs a rebase. We can help if you like.

sergioisidoro:
Ok, if I understood correctly, I'll:

* Rebase
* Move the dep to temporarily dependencies
* Lint so that the deps are ordered alphabetically
* Double check that this does not apply to 2.x and if so, remove this from caveman templates

Got it!

Dockerfile-debian.template
```diff
@@ -20,6 +20,7 @@ ENV PYTHON_VERSION %%PLACEHOLDER%%
 
 RUN set -ex \
 	\
+	&& apt-get update && apt-get install -y --no-install-recommends libbluetooth-dev \
```
sergioisidoro: Please note that I did not make anything regarding the  `savedAptMark` and `apt-mark showmanual`. 

I have very little clue about its meaning, but it seemed to be in the slim image.

sergioisidoro:
Double checked that `AF_BLUETOOTH` is never mentioned in the python 2.x docs for the sockets family, so I removed the deps from the caveman templates

Dockerfile-debian.template
tianon: Sorry for not being more clear -- this one was fine before (with `libbluetooth-dev` up above `libssl-dev`'s comment line above), ala:

```diff
diff --git a/Dockerfile-debian.template b/Dockerfile-debian.template
index 4d56b25..8a14f49 100644
--- a/Dockerfile-debian.template
+++ b/Dockerfile-debian.template
@@ -9,6 +9,7 @@ ENV LANG C.UTF-8
 
 # extra dependencies (over what buildpack-deps already includes)
 RUN apt-get update && apt-get install -y --no-install-recommends \
+		libbluetooth-dev \
 # Python 3.4 on Stretch+ needs to use an older version of "libssl-dev" (these lines both get removed for every other combination)
 		libssl-dev \
 		tk-dev \
```

Happy to take over and make this change directly to your branch if you'd rather. :+1:

Thanks for updating this! :metal:

sergioisidoro: Oh, my fault. Now that I re-read the comment it was clear that is was only referring to the slim image, not all Deb. 😅

I can also change it, since it's 5 min, but feel free to push to branch if you want :)
I'll stop force pushing.

sergioisidoro: Done.

I force pushed one more time to keep the update.sh as the last commit.
I don't know what is the merge strategy, but maybe squash and merge would be beneficial here.

Also, does that libssl sill make sense? 
3.4 is not being rendered, so libssl never makes it to any of the images.

tianon: Oh, nice catch!  Maybe we should clean that out separately from this though. :+1:

PR #446, opened 2020-01-01
Add windowsservercore-1809 to AppVeyor

jairov4:
current update.sh is lacking the latest windowsservercore version which is 1909

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,16 +1,31 @@
 version: build-{build}.{branch}
-image: Visual Studio 2017
 
 environment:
   matrix:
     - version: 3.9-rc
       variant: windowsservercore-ltsc2016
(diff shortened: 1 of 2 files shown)


tianon:
See https://github.com/docker-library/golang/pull/289#issuecomment-509841894 (which also applies to 1909):

> We don't currently have a place for these to build, unfortunately (we currently use a Windows Server 2019 VM which is nice because it can actually be updated properly, but has the downside that semi-annual releases like 1903 don't get included). :disappointed:

Also, `ltsc2019` is a rolling tag which is currently the same as `1809` (so we have opted not to explicitly build any `FROM ...:ltsc2019` images, preferring instead to follow the individual kernel compatibility tracks directly).

tianon:
AppVeyor is green -- thanks for the heads up on `APPVEYOR_BUILD_WORKER_IMAGE`! :metal:

PR #497, opened 2020-06-16
Updated README.md

swapnanildutta:
Corrected some grammatical errors.


tianon:
These are all in the backticked code syntax because they're not referring to "Python, the language" but rather "`python`, an image on Docker Hub" (and the entire file is generated via https://github.com/docker-library/docs/blob/1a91d48fdb0a6eded8adfef1709d76d51ba1d8dd/generate-repo-stub-readme.sh).

PR #502, opened 2020-07-03
Fix #501: add -fno-semantic-interposition and --with-lto

### LINKED ISSUE #501: Disable semantic interposition?
Red Hat recently released an article about how they improved Python's performance by 30% in Python 3.8 by disabling semantic interposition. Article: https://developers.redhat.com/blog/2020/06/25/red-hat-enterprise-linux-8-2-brings-faster-python-3-8-run-speeds/

Looks like they added the `-fno-semantic-interposition` flag to GCC. 

Would this option make sense for this image?

blopker:
PR with the proposed changes from #501.

### DIFF
--- a/3.5/alpine3.11/Dockerfile
+++ b/3.5/alpine3.11/Dockerfile
@@ -71,14 +71,15 @@ RUN set -ex \
 		--enable-optimizations \
 		--enable-option-checking=fatal \
 		--enable-shared \
+		--with-lto \
 		--with-system-expat \
 		--with-system-ffi \
 		--without-ensurepip \
 	&& make -j "$(nproc)" \
 # set thread stack size to 1MB so we don't segfault before we hit sys.getrecursionlimit()
 # https://github.com/alpinelinux/aports/commit/2026e1259422d4e0cf92391ca2d3844356c649d0
-		EXTRA_CFLAGS="-DTHREAD_STACK_SIZE=0x100000" \
-		LDFLAGS="-Wl,--strip-all" \
+		EXTRA_CFLAGS="-DTHREAD_STACK_SIZE=0x100000 -fno-semantic-interposition" \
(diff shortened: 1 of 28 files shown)


itamarst:
Just wanted to ping maintainers about this. Given this is an image that is probably installed on tens of thousands, if not more, machines, this tiny change can save a lot of people a lot of money, not to mention reduce a lot of carbon emissions.

emhagman:
I'd like to +1 back-porting this to < 3.10. We currently use the latest `security` Python release at work which is `3.8` and can't get the benefits of this without switching to a different image.

blopker:
I'm curious, what's a `security` release? This seems to imply that newer versions of Python are less secure than older ones.

emhagman:
@blopker I just mean the releases of Python not still in `bugfix` mode, listed on their website. It doesn't mean they are less secure, but simple more "stable" from a production software standpoint. If 3.8 has bugs, they won't fix them but they're not going to change either.

CleanShot 2021-11-20 at 10 55 46@2x

blopker:
Fair enough. I don't know anything about the stuff you work on, but it seems to me that the security status only means it's not getting any non-security changes and not that it's more reliable. Django's LTS releases include data loss issues for instance. According to https://devguide.python.org/devcycle/ the change in maintenance status is actually kinda of arbitrary, decided by one release manager without any hard criteria other than one bugfix release.

Anyway, my point is updating to 3.10 shouldn't sacrifice stability any more than updating to another 'security' release. I get that there are other compatibility considerations when jumping 2 releases (I personally updated a million+ line codebase from 2.7 to 3.8, bleh). However, that may be worth it if the project is performance critical. Plus you get all the new goodies in 3.10, which does include even more performance enhancements than just this one.

Maybe I'm weird, but I generally like challenging company policies :)

PR #522, opened 2020-09-30
Update update.sh

harshitkapoor00:


tianon:

Thanks!

PR #536, opened 2020-10-06
Create Python Program to Shuffle Deck of Cards

AyushSharma-code:
In the program, we used the product() function in itertools module to create a deck of cards. This function performs the Cartesian product of the two sequences.

The two sequences are numbers from 1 to 13 and the four suits. So, altogether we have 13 * 4 = 52 items in the deck with each card as a tuple. For example,

deck[0] = (1, 'Spade')

Our deck is ordered, so we shuffle it using the function shuffle() in random module.

Finally, we draw the first five cards and display it to the user. We will get different output each time you run this program as shown in our two outputs.

Here we have used the standard modules itertools and random that comes with Python.

### DIFF
--- a/Python Program to Shuffle Deck of Cards
+++ b/Python Program to Shuffle Deck of Cards
@@ -1,0 +1,15 @@
+# Python program to shuffle a deck of card
+
+# importing modules
+import itertools, random
+
+# make a deck of cards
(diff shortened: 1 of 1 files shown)


tianon:
I'm sorry, but we are not participating in Hacktoberfest, and this is definitely not an appropriate place to submit this.  I'd suggest creating your own personal repository and hosting your Python programs there.

PR #538, opened 2020-10-06
Add 3.10.0a1

tianon:
See https://pythoninsider.blogspot.com/2020/10/python-390-is-now-available-and-you-can.html


tianon:
GitHub Actions is having a bit of a flake (https://www.githubstatus.com/incidents/t34640ccrmfs), but the 3.10 builds are all green :innocent:

PR #557, opened 2020-10-28
Set PYTHONIOENCODING for python-windowsservercore

JaehunYoon:
Related #147 

I'm using `python:3.8-windowsservercore-1809` image.
When I try to print a Unicode character in docker container without --tty, I get an error like this:

```
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-2: character maps to <undefined>
```

### DIFF
--- a/Dockerfile-windowsservercore.template
+++ b/Dockerfile-windowsservercore.template
@@ -2,6 +2,9 @@ FROM mcr.microsoft.com/windows/servercore:%%PLACEHOLDER%%
 
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
+# https://github.com/docker-library/python/pull/557
+ENV PYTHONIOENCODING UTF-8
+
 ENV PYTHON_VERSION %%PLACEHOLDER%%
 ENV PYTHON_RELEASE %%PLACEHOLDER%%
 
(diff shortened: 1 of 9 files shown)


Dockerfile-windowsservercore.template
```diff
@@ -4,6 +4,7 @@ SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPref
 
 ENV PYTHON_VERSION %%PLACEHOLDER%%
 ENV PYTHON_RELEASE %%PLACEHOLDER%%
+ENV PYTHONIOENCODING UTF-8
```
tianon, line 7: This should move up after `SHELL` but before `PYTHON_VERSION`, and ought to include a link (similar to what we did in #278):

```suggestion

# https://github.com/docker-library/python/pull/557
ENV PYTHONIOENCODING UTF-8

ENV PYTHON_VERSION %%PLACEHOLDER%%
ENV PYTHON_RELEASE %%PLACEHOLDER%%
```

tianon (changes requested):
Ahh, we added this for Python 2 in #278, but not for Python 3 since it was sourced automatically from `LANG`, which Windows doesn't use. :sweat_smile:

Dockerfile-windowsservercore.template
JaehunYoon: Thanks, I modified it.

tianon (approved):
LGTM, thanks!

cc @yosifkit

PR #570, opened 2021-01-08
Removing LANG env var from 3.10-rc and templates

henryh9n:
Since `ENV LANG C.UTF-8` is apparently [no longer necessary](https://bugs.python.org/issue19846), removed those from 3.10-rc and templates not to cause confusion.

### DIFF
--- a/3.10/alpine3.14/Dockerfile
+++ b/3.10/alpine3.14/Dockerfile
@@ -9,10 +9,6 @@ FROM alpine:3.14
 # ensure local python is preferred over distribution python
 ENV PATH /usr/local/bin:$PATH
 
-# http://bugs.python.org/issue19846
-# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
-ENV LANG C.UTF-8
-
 # runtime dependencies
 RUN set -eux; \
 	apk add --no-cache \
--- a/3.10/alpine3.15/Dockerfile
+++ b/3.10/alpine3.15/Dockerfile
@@ -9,10 +9,6 @@ FROM alpine:3.15
 # ensure local python is preferred over distribution python
 ENV PATH /usr/local/bin:$PATH
 
-# http://bugs.python.org/issue19846
-# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
-ENV LANG C.UTF-8
-
 # runtime dependencies
 RUN set -eux; \
 	apk add --no-cache \
--- a/3.10/bullseye/Dockerfile
+++ b/3.10/bullseye/Dockerfile
@@ -9,10 +9,6 @@ FROM buildpack-deps:bullseye
 # ensure local python is preferred over distribution python
 ENV PATH /usr/local/bin:$PATH
 
-# http://bugs.python.org/issue19846
-# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
-ENV LANG C.UTF-8
-
 # runtime dependencies
 RUN set -eux; \
 	apt-get update; \
--- a/3.10/buster/Dockerfile
+++ b/3.10/buster/Dockerfile
@@ -9,10 +9,6 @@ FROM buildpack-deps:buster
 # ensure local python is preferred over distribution python
 ENV PATH /usr/local/bin:$PATH
 
-# http://bugs.python.org/issue19846
-# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
-ENV LANG C.UTF-8
-
 # runtime dependencies
 RUN set -eux; \
 	apt-get update; \
--- a/3.10/slim-bullseye/Dockerfile
+++ b/3.10/slim-bullseye/Dockerfile
@@ -9,10 +9,6 @@ FROM debian:bullseye-slim
 # ensure local python is preferred over distribution python
 ENV PATH /usr/local/bin:$PATH
 
-# http://bugs.python.org/issue19846
-# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
-ENV LANG C.UTF-8
-
 # runtime dependencies
 RUN set -eux; \
 	apt-get update; \
--- a/3.10/slim-buster/Dockerfile
+++ b/3.10/slim-buster/Dockerfile
@@ -9,10 +9,6 @@ FROM debian:buster-slim
 # ensure local python is preferred over distribution python
 ENV PATH /usr/local/bin:$PATH
 
-# http://bugs.python.org/issue19846
-# > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
-ENV LANG C.UTF-8
-
 # runtime dependencies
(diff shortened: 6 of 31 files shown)


tianon:
Unfortunately this won't work as-is (post-merge we'll have @docker-library-bot immediately apply the template change across all image variants/versions), but this is really interesting!

From my reading of that issue, it sounds like it's been unnecessary since Python 3.7? :astonished:

(It's too bad 3.6 isn't quite EOL yet or this would be really easy for us to remove! :smile:)

henryh9n:
> Unfortunately this won't work as-is (post-merge we'll have @docker-library-bot immediately apply the template change across all image variants/versions), but this is really interesting!
> 
> From my reading of that issue, it sounds like it's been unnecessary since Python 3.7? 😲
> 
> (It's too bad 3.6 isn't quite EOL yet or this would be really easy for us to remove! 😄)

Yeah as far as I can see it is obsolete for at least version 3.7 (there are some discussions for earlier versions, but I'm not sure :)) So what should we do? Will changing only for 3.10-rc work? (ofc later removing it in templates) or should we wait for 3.6's end of the life?

edmorley:
Hi! Now that Python 3.6 is EOL, is this worth reconsidering? :-)

henryh9n:
@edmorley Thanks for the reminder. Done!
cc: @tianon

henryh9n:
@edmorley @tianon Checks look good. Should we merge?

tianon:
Yes, thank you!  Was waiting for the checks but hadn't gotten back to it yet. :pray:

prograwertyk:
We are using python:3.9-slim-buster in our projects and after switched to `python:3.9-slim-buster@sha256:9fd7e4aaa6aae1d390d09400cad6c14508f5a2cc543987f162c9d3da753e2879` (image that has removed  [`ENV LANG C.UTF-8`](https://github.com/docker-library/python/pull/570/files#diff-695faa57812bc7d36b06c958acc3c7feb385b2692b399b47fd5b76820bb98b92) from Dockerfile) our apps was generating a massive amount of `UnicodeEncodeErrors`.

After switching back to `python:3.9-slim-buster@sha256:59356005f0cd210ab8f81f08f2aa78528c6049a500cd241f06e209b0dff09515` problems gone.

@henryh9n Let me know if you need some more info.

sagata1999:
**python:3.8** either falls apart to UnicodeEncodeErrors in cause of 'ascii' not knowing about utf-8 characters (anyway I'm thinking so). But, when we settled up `ENV LANG=C.UTF-8` issue seems to disappear...

@henryh9n

tianon:
Can we come up with a tiny reproducer? 😇

mauriziomocci:
We are using python:3.9-slim-buster in our project. After  a "friday afternoon deploy" before holidays our app started generating lots of UnicodeEncodeErrors. We've lost 3 hours investigating the issue :(

I attach a useful image that describes the problem:

grapo:
In our case there is a problem with uWSGI when using latest python:3.9-slim-buster
Uwsgi sets stderr and stdout  (https://github.com/unbit/uwsgi/blob/master/plugins/python/python_plugin.c#L598) based on `locale.getpreferredencoding()` 

We place `print(locale.getlocale(locale.LC_CTYPE), locale.getpreferredencoding(), sys.flags.utf8_mode)` in code and run uwsgi:
```
$ uwsgi --ini /uwsgi.ini
...
(None, None) ANSI_X3.4-1968 0
...

$ LANG=C.UTF-8 uwsgi --ini /uwsgi.ini
...
('en_US', 'UTF-8') UTF-8 0
...
```
If encoding is set to `ANSI_X3.4-1968` then any string with nonascii character send to logger causes  `UnicodeEncodeError`

Currently I have no idea why uwsgi don't use system locale

lambda:
We ran into issues after updating as well; scripts which use Click will [refuse to run without a Unicode locale](https://click.palletsprojects.com/en/5.x/python3/#python-3-surrogate-handling).

Changing the locale like this has some pretty wide-ranging effects, I think it would be better to revert this change and leave the locale as `C.UTF-8`. This is a breaking change for a number of use cases, and anyone who needed a different locale would have already overridden this in their use case; it seems like the change causes more problems than it solves.

lambda:
Huh. An attempt to create a minimal example of the failure using Click is not showing the issue. I may need to dig a bit to figure out a minimal set of steps to reproduce.

drzraf:
This **update broke all my CI**, not only because its used `python:alpine` but all available images are affected making very hard to track it down.

Ex:
`docker run -ti python:alpine3.14 sh -c "pip3 install --user babel && python -c 'from babel.core import default_locale, Locale; print(Locale.parse(default_locale(\"LC_NUMERIC\")))'"`

- **Before**: `C.UTF-8`
- **Now**: `None`

And consequently:
`docker run -ti python:alpine3.14  sh -c "pip3 install --user babel && python3 -c 'from babel.numbers import parse_decimal; parse_decimal(\"32.2\");'"`

**Before**: `works`
**Now**: 
```
  File "/root/.local/lib/python3.10/site-packages/babel/numbers.py", line 729, in parse_decimal
    group_symbol = get_group_symbol(locale)
  File "/root/.local/lib/python3.10/site-packages/babel/numbers.py", line 332, in get_group_symbol
    return Locale.parse(locale).number_symbols.get('group', u',')
AttributeError: 'NoneType' object has no attribute 'number_symbols'
```

I highly suggest to roll this back thinking about all the current containers breaking in strange way without any clear and easily identifiable reason.

MT-Cash:
This also broke my CI. I don't completely grasp it, but opening files with German umlauts in the path doesn't work anymore:

```
with open("Ausführung.pdf", "rb") as fp:  # shortened a bit
    data = fp.read()
```

suddenly gives 

> 'ascii' codec can't encode character '\xfc' in position 44: ordinal not in range(128)

tianon:
Reverted via https://github.com/docker-library/python/pull/719 :disappointed: :see_no_evil:

edmorley:
> This **update broke all my CI**, not only because it changed the basic behavior of `python:alpine`

@drzraf I can reproduce the failure in your post.

However, some observations:
- The expected result in your first example seems to be incorrect? It says `Before: C.UTF-8` but I got `en_US_POSIX` for it.
- The particular issue you saw with babel only seems to affect alpine (unlike the other issues which were reported on other distros too). For example it works on Debian on multiple Python version and Debian version variants, eg:
   ```
   docker run -ti python:3.10 sh -c "unset LANG && pip3 install --user babel && python3 -c 'from babel.numbers import parse_decimal; parse_decimal(\"32.2\");'"
   docker run -ti python:3.7-buster sh -c "unset LANG && pip3 install --user babel && python3 -c 'from babel.numbers import parse_decimal; parse_decimal(\"32.2\");'"
   ```

edmorley:
Whilst #719 resolved this issue (by reverting the `LANG` removal), the code comment explaining why `LANG` is needed still references the already-fixed Python bug:
https://github.com/docker-library/python/blob/1cf43e70e45843c70909a5f914c3c6d0f85fc200/Dockerfile-linux.template#L22-L24

It feels like figuring out what broke here, reporting it upstream if appropriate (given the bugs PEPs I've seen seem to suggest this should be resolved already?), and updating the code comment to indicate why `LANG` is still needed might help prevent future similar breakages (plus allow for cleanup of `LANG` once appropriate).

PR #572, opened 2021-01-15
Add alpine3.13 to 3.10-rc

cschramm:
automake 1.16.3 has a fix to work with Python 3.10 and alpine 3.13 is currently the only viable base image that ships it.


ElPincheTopo:
@cschramm Is it worth adding alpine3.13 to the other versions of python as part of this pull request?

cschramm:
Shouldn't hurt. -> #574

PR #574, opened 2021-01-22
Add alpine3.13

cschramm:
automake 1.16.3 has a fix to work with Python 3.10 and alpine 3.13 is currently the only viable base image that ships it.


update.sh
```diff
@@ -132,7 +132,7 @@ for version in "${versions[@]}"; do
 	echo "$version: $fullVersion"
 
 	for v in \
-		alpine{3.11,3.12} \
+		alpine{3.11,3.12,3.13} \
```
tianon, line 135: I think we need to decide if we want to also drop Alpine 3.11 here (and keep in line with what we've done in Ruby and PHP and only support two at a time).

@yosifkit thoughts?

yosifkit: Agreed, we can only support two versions of Alpine at a time. Added commit for that and updating `./generate-stackbrew-library.sh`

yosifkit (approved):

PR #581, opened 2021-02-21
Create buildspec.yml

raghavan483:
added new buildspec


tianon:
I think this was opened in the wrong place.

Issue #61, opened 2015-08-31
Any plan to have support for precise


ashrafuzzaman:
Hi guys,
Do you have any plan to support legacy system like `ubuntu precise`, as a lot of existing production environment is still built on `precise`. If you like I would love to come up with a pull request.

yosifkit:
We actually only have the python images based on `debian`, so `ubuntu` images are not something we have looked to do.  Are there specific python libraries that you need from `precise`?  From my understanding, running python in a container means that I don't have to care what "OS" it is packaged inside of, be it debian, ubuntu, or centos, it will just work.  Using the non-slim versions I need only to `pip install` my dependencies and my python application can be up and running.

ashrafuzzaman:
Sorry for taking so much time to reply. In ideal world we should be able to install packages using pip. But that is not the case. For an example installing [pil](http://codeinthehole.com/writing/how-to-install-pil-on-64-bit-ubuntu-1204/) is different in different os.

tianon:
I think https://github.com/docker-library/python/pull/65 and https://github.com/docker-library/python/issues/104 are relevant discussions along a similar vein to this one.

I can say for sure that we won't be adding a `precise` variant unless there is a _really_ strong justification for doing so.

eromoe:
Many software installation are different by platfom. I'd like to use python docker base on ubuntu, because I only know installation steps on ubuntu.

yosifkit:
Thank you for your contribution and time.

Unfortunately this is not something we want to support and maintain. For most users, just having "python" at a specific version fulfills all their requirements. They don't need to care about the underlying distribution it is running within. If they are running their own application, in python for example, and they are unable to run it in our Debian, Alpine, or Windows Server Core based python images, then they are free to install python and their required dependencies within CentOS, Ubuntu, or other base distribution image.

For many images here are the variants that we maintain with short descriptions and why we have them:

 - `debian`
   - what: the default image that is based on [Debian](https://www.debian.org/) but doesn't usually say `debian` in the name
   - why: the smallest mainline distro using glibc; stable set of libraries provided though `apt`; we have familiarity with Debian; it is similar enough to Ubuntu for most users
 - `slim`:
   - what: a smaller version of the Debian based image; to make it really easy to use, many of the language images include commonly used libraries and development headers; this variant doesn't install them
 - `stretch` (or `wheezy`)
   - what: Debian based image built on the next (or previous) stable [release of Debian](https://www.debian.org/releases/)
[...]

PR #582, opened 2021-02-22
Add support for Ubuntu Bionic base image

christianbundy:
Hi! This PR adds support for the Ubuntu Bionic base image, which is very similar to Debian Buster with a small change to IPv6 support when running `gpg --recv-keys`. Having this as an official image would add support for folks who are using Ubuntu-specific features but still want to continue running the official Docker Python image.

Thanks!

### DIFF
--- a/Dockerfile-bionic.template
+++ b/Dockerfile-bionic.template
@@ -1,0 +1,130 @@
+FROM buildpack-deps:%%PLACEHOLDER%%
+
+# ensure local python is preferred over distribution python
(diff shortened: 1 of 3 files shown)


tianon:
See #61, #65, and #104 which are all closely related discussions (see especially https://github.com/docker-library/python/issues/61#issuecomment-282450114).

christianbundy:
Thanks for the info, I hadn't seen those.

PR #586, opened 2021-03-02
Catch errors in windows on python installer failure

yosifkit:
This should hopefully help us catch why `ltsc2016` is failing in https://github.com/docker-library/official-images/pull/9722


tianon:
Looks like this is our failure, and from what I can tell by searching around I think it actually comes from the VC redistributable that must be embedded in the Python installer (and the error is supposedly permission denied, which is odd):

```
[1488:1478][2021-03-02T18:07:38]e000: Error 0x80070005: Failed to get provider key bundle id.
[1488:1478][2021-03-02T18:07:38]e000: Error 0x80070005: Failed to detect provider key bundle id.
[1488:1478][2021-03-02T18:07:38]i199: Detect complete, result: 0x80070005
```

tianon:
(Just a rebase for a fresh run of the builds.)

tianon (approved):

PR #588, opened 2021-03-06
update pypa/get-pip branch from master -> main

westonsteimel:
It looks like https://github.com/pypa/get-pip/ updated their default branch from `master` to `main` and deleted the master branch, so this updates the update script accordingly so that it doesn't get a 404


yosifkit:
Thanks!

PR #589, opened 2021-03-06
Add link to official images by python upstream

hginzel:


tianon:
I'm OK with linking to these, but IMO the description on https://hub.docker.com/_/python is probably a more appropriate place (esp. since this README file is autogenerated from a boilerplate template for all the repositories in https://github.com/docker-library).

In linking to them, I think we also need to describe the caveats of what they are (multiple versions of Python in a single image), so users don't think they're general purpose replacements for what we maintain here. :sweat_smile:

(See also https://github.com/docker-library/python/issues/373#issuecomment-461116541 for where these have been discussed previously.)

Issue #595, opened 2021-03-30
CVE-2019-18276 requires debian bullseye


bneijt:
Security issue https://security-tracker.debian.org/tracker/CVE-2019-18276 is not a big deal for anybody running a docker image as root, but still nice to be able to avoid.

Currently it requires debian bullseye, which should be easy to add to the python images as an option. Please note that Debian bullseye has not been officially released yet and no official date has been set ATM.

Hopefully this issue can help track progress on this CVE.

wglambert:
The Debian security team notes that it has a "negligible security impact" and that the urgency is "unimportant"

Some more information on this particular CVE so you can determine how applicable it is to your Python usecase https://amriunix.com/post/cve-2019-18276-suidbash/

See also https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves
And https://github.com/docker-library/openjdk/issues/449#issuecomment-763027011, https://github.com/docker-library/postgres/issues/286#issuecomment-302512767 docker-library/openjdk#161, docker-library/openjdk#112, docker-library/postgres#286, docker-library/drupal#84, docker-library/official-images#2740, docker-library/ruby#117, docker-library/ruby#94, docker-library/python#152, docker-library/php#242, docker-library/buildpack-deps#46, docker-library/openjdk#185.

PR #596, opened 2021-03-30
Add debian bullseye

bneijt:
See issue https://github.com/docker-library/python/issues/595 on CVE-2019-18276

### DIFF
--- a/3.10-rc/bullseye/Dockerfile
+++ b/3.10-rc/bullseye/Dockerfile
@@ -1,0 +1,97 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "update.sh"
+#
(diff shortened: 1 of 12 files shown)


tianon:
Mirroring my comment from https://github.com/docker-library/official-images/pull/9901#issuecomment-812201729, which is very relevant:

> Putting my Debian hat on, I'd add a warning here that Bullseye is _not_ officially security-supported (yet -- until release), and isn't quite in [full freeze](https://release.debian.org/bullseye/freeze_policy.html#full) yet, so while it's probably a pretty stable base (and I actively run it on several of my host systems), it's still technically a little bit of a moving target. :smile:

We'll reconsider adding Bullseye after it's officially released.

PR #599, opened 2021-03-31
Switch to "keys.openpgp.org"

tianon:
The keys.openpgp.org service is much more reliable than the SKS network, and every current release manager appears to have confirmed their email address over there.


yosifkit:
Any thoughts on dropping `pgp-happy-eyeballs` for a single repo? Or perhaps we wait to make it optional once we have more repos not needing it.

(https://github.com/docker-library/bashbrew/blob/9df92226ae134cac637af04e6d466b82f0508c84/scripts/github-actions/generate.sh#L167-L172)

tianon:
Yeah, not a bad idea -- that could probably even do some basic detection of whether it's necessary, but I guess it would be better to wait until we get more things that don't need it at all to figure out the "right" approach (who knows, maybe some future release of Python will have a release manager who doesn't want to verify on keys.openpgp.org :grimacing:).

yosifkit (approved):

PR #602, opened 2021-04-05
Update to Python 3.9.4 (hotfix)

Tenzer:
https://blog.python.org/2021/04/python-394-hotfix-is-now-available.html


tianon:
Thanks!

noelleleigh:
@tianon [3.10.0a6-windowsservercore-ltsc2016 seems to be stuck](https://github.com/docker-library/python/runs/2270520369) on b147198

tianon:
Good catch; I've killed/restarted it now (although for getting the builds out, it doesn't actually matter -- that's what https://github.com/docker-library/official-images/pull/9925 is :+1:)

PR #617, opened 2021-05-20
Fix "SSLError"

SINOPW:
without "apt-get install libssl-dev" before compiling python:
pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.

### DIFF
--- a/3.8/buster/Dockerfile
+++ b/3.8/buster/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		libbluetooth-dev \
 		tk-dev \
 		uuid-dev \
+		libssl-dev \
 	&& rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


tianon:
I'm 2x confused here. :sweat_smile:

1. `buildpack-deps:buster` already includes `libssl-dev`:

    ```console
    $ docker run --rm buildpack-deps:buster dpkg -l | grep libssl-dev
    ii  libssl-dev:amd64                   1.1.1d-0+deb10u6             amd64        Secure Sockets Layer toolkit - development files
    ```

2. we have a test which verifies that SSL works in all variants of this image:

    https://github.com/docker-library/official-images/blob/ef61ecd87c552a2ee01fc33b5105f3d33fc2b242/test/tests/python-pip-requests-ssl/container.py

    (That test is auto-run against all versions in the CI here and over in the official-images repository on every update PR)

PR #623, opened 2021-06-29
Bump Python versions

Tenzer:
New versions are out of Python 3.9, 3.8, 3.7 and 3.6: https://pythoninsider.blogspot.com/2021/06/python-396-3811-3711-and-3614-are-now.html.


Tenzer:
Oh, I just noticed the Windows builds for 3.8.11 is failing because there are no longer binary releases available for it. The blog post I linked to also states:

> Python 3.8 isn’t receiving regular bugfixes anymore, and binary installers are no longer provided for it. Python 3.8.10 was the last full bugfix release of Python 3.8 with binary installers.

I can see there aren't any Windows versions for Python 3.6 and 3.7. Would that mean the right thing to do here is to remove the Windows images for Python 3.8 as well?

tianon:
Yep, that's correct -- @docker-library-bot has been working on making this same change all night, but it keeps timing out because there's so many versions to build. :joy:

Tenzer:
Alright, I have removed the 3.8 Windows Docker images in 6166798.

tianon:
Looks good, thanks! :+1:

Issue #630, opened 2021-08-06
docker build fails for alpine when running behind a proxy


kodefoundry:
docker build is failing for alpine versions when run behind a proxy due to failure of wget command.
This is due to https://gitlab.alpinelinux.org/alpine/aports/-/issues/10446

The proposed solution is to override busybox wget with wget from apk.
'apk add wget'

kodefoundry:
The only difference I see is 
apk wget version         --> GNU Wget 1.21.1 built on linux-musl
busybox wget version --> BusyBox v1.33.1 () multi-call binary

tianon:
We don't build behind a proxy, so we haven't run into this (and if we did, we'd likely use transparent proxying rather than explicit environment variables).

I'm curious why you need to build from source instead of using the published images as-is?

kodefoundry:
Hi Tianon,
Thank you for your kind response.
To answer your question, "why you need to build from source instead of using the published images as-is?"
In our organisation, our security team provides hardened base OS images that we are supposed to use for all our containers.
I maintain a couple of Python based repo's and it was a pain for us to upgrade to latest Python versions, So I created a Docker file to build Python from source as done in the official python image but using our hardened OS as base image.

I actually ran into the proxy problem in 2 places one is with wget and another is with the line where GPG key is downloaded. Which I figured out after I raised this issue.

I would like to know if you guys would like to adopt this change, in which case I can go and make changes as required in other maintained versions also, else you can reject the PR

Thanks
Biju

tianon:
Interesting, thanks for explaining.

I'm afraid as it stands, we'd be more -1 on such a change since we wouldn't be testing that use case at all, and would thus be very likely to break it again in the future.  I'd suggest that whatever process you're using to consume from our repository could be updated to include something like `RUN apk add --no-cache wget` directly after the `FROM` line (which you likely have to adjust anyways).

PR #631, opened 2021-08-06
docker build fails for alpine running behind proxy

kodefoundry:
https://github.com/docker-library/python/issues/630

This is because of issue with busybox linux that comes default 
see https://gitlab.alpinelinux.org/alpine/aports/-/issues/10446

The proposed solution is to override busybox wget with wget from apk.
'apk add wget'


tianon:
Closing per discussion in #630

PR #633, opened 2021-08-12
Add images based on Debian 11 (Bullseye)

michael-k:
The stable release of Debian 11 (Bullseye) is expected on 2021-08-14, see https://lists.debian.org/debian-devel-announce/2021/07/msg00003.html

The base images `buildpack-deps:bullseye` and `debian:bullseye-slim` are available on Docker Hub.


michael-k:
> after the release is officially released

That's what I expected :)

> drop all `stretch` variants

Done

generate-stackbrew-library.sh
```diff
@@ -6,9 +6,9 @@ declare -A aliases=(
 	[3.9]='3 latest'
 )
 
-defaultDebianSuite='buster' # TODO buster
```
tianon, line 9: :joy:

tianon (approved):
LGTM :+1:

generate-stackbrew-library.sh
zacharylyonstx: This caused me a lot of frustration over the past day, can I get more context on the change than "😂 "? And to anyone who uses the `FROM python:3.8` on your dockerfile and you are wondering why you are magically having issues installing nodejs in your build, just change your dockerfile to `FROM python:3.8-buster`

tianon: The :joy: is specifically because we had `buster # TODO buster`, which is silly (when it got updated to buster in https://github.com/docker-library/python/commit/2a11f610a56ff3c0f0157790dde940894fad7a1a we forgot to remove the comment).

The update of the default variant was our standard procedure for these repositories and distribution updates like this, and the reason they provide more specific aliases for folks to continue to pin to `buster`.

zacharylyonstx: Forgive my frustration, having a bit of a crash course in how these images are managed. I find my first comment a little funny now since this is literally the point of the PR. That being said, it just feels weird that so much of the documentation around setting up the docker environment would be around using generic images like Python:3.8, where at any given time, the OS version can change.

This was my mistake for not pinning the version in the first place, thanks for the clarification!

blackdog308:
#636 Since the update I've been unable to install the drivers for MSSQL I thought I had broken something, but I'm guessing it could be these changes?

michael-k:
> I'm guessing it could be these changes

Only if you're using `python:lastest`.  But I hope you're not doing that.

blackdog308:
> > I'm guessing it could be these changes
> 
> Only if you're using `python:latest` or any other tag that doesn't mention the OS version (eg. `python:3.9` instead of `python:3.9-buster` or `python:3.9-bullseye`). But I hope you're not doing that.

Thanks @michael-k 

Learnt something new today. Appreciate the help.

PR #647, opened 2021-09-07
Pin pip to 21.2.x and setuptools to 57.x

### LINKED ISSUE #365: How about updating setuptools, wheel, pip and installing setuptools-scm from git?
Rationale: the stuff mentioned has a very long release cycle. It takes a while before features and fixes available in git are landed into release. I have been using the bleeding edge stuff from git in CI (installed manually upon `python:latest`) for a long time and experienced no problems with it, only benefits. So I think it may make sense to include this stuff into the images to ship it by default.

What does the community think about it?

tianon:
Closes #365

### DIFF
--- a/3.10-rc/alpine3.13/Dockerfile
+++ b/3.10-rc/alpine3.13/Dockerfile
@@ -115,6 +115,8 @@ RUN cd /usr/local/bin \
 
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
 ENV PYTHON_PIP_VERSION 21.2.4
+# https://github.com/docker-library/python/issues/365
+ENV PYTHON_SETUPTOOLS_VERSION 57.5.0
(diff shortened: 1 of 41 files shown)


tianon:
86c04e958f4af942cd95ce805990c2270c090f90 was green, https://github.com/docker-library/python/compare/86c04e958f4af942cd95ce805990c2270c090f90..bb68424de76756a2d3dc817f87b1f8640112461f is just fixing a typo :innocent:

tianon:
(see https://github.com/pypa/setuptools/blob/b254ea77cd8328ab6e9fbe84db6718bcf6b014d4/CHANGES.rst#v5800 for the breaking changes that have prompted this to finally happen...)

tianon:
(see also https://github.com/docker-library/python/pull/194 for where we had version numbers for these previously, but removed the weirder logic around them)

yosifkit (approved):

PR #653, opened 2021-10-05
Update to 3.10.0 (GA)

timgates42:


tianon:
Nice, thanks!

haakenlid:
The directory name should probably be renamed from "3.10-rc" to "3.10"

yosifkit:
> The directory name should probably be renamed from "3.10-rc" to "3.10"

It is!
> 3.10-rc/alpine3.13/Dockerfile → 3.10/alpine3.13/Dockerfile

PR #670, opened 2021-11-24
Add Alpine 3.15

darktohka:
Alpine Linux 3.15 was released today:
https://alpinelinux.org/posts/Alpine-3.15.0-released.html

This pull request adds Alpine 3.15 support, while removing Alpine 3.13 builds. This is the same action that was taken after Alpine 3.14's release.

Build matrix:
3.6, 3.7, 3.8, 3.9, 3.10: builds for Alpine 3.14 and Alpine 3.15
3.11-rc: builds for Alpine 3.15


tianon (approved):
Nice, thank you!

PR #673, opened 2021-12-07
Upgrade to Python 3.10.1

Tenzer:
https://www.python.org/downloads/release/python-3101/


yosifkit:
Looks like the bot got it: https://github.com/docker-library/python/commit/db32c78 and the build PR is getting ready: https://github.com/docker-library/official-images/pull/11461

PR #679, opened 2022-01-15
Update to 3.10.2, 3.9.10, 3.11.0a4

tianon:
This is a temporary hack to account for the Windows installer issues described in https://pythoninsider.blogspot.com/2022/01/python-3102-3910-and-3110a4-are-now.html

I'm opening this as a reminder of the branch I created -- not because I think we should actually merge it.  Hopefully next week we'll have Windows installers and this whole workaround will become a moot point (and we can then just delete the branch).

I created a separate branch so that `update.sh` doesn't have to change for this (temporary) hack, and so that @docker-library-bot doesn't revert the change right away.

(See https://github.com/docker-library/official-images/pull/11684 for the update PR.)


ulgens:
Is there anything we can do to speed up this PR? Heroku thinks using 3.9.9 is a security flaw, but there is no official image yet to fix it.

tianon:
This PR is just a placeholder and isn't going to be merged -- the PR we want/need to merge is https://github.com/docker-library/official-images/pull/11684.

tianon:
Oh, it looks like the Windows issues described are now resolved!, so we can close this entirely!

PR #483, opened 2020-05-13
add LDFLAGS for optimizing image size

autoferrit:
This is based on #448 by @lubo 

I ran this locally with the change and it worked out well. I build the `3.8-alpine3.11` image (tagging it as just `py3` here) and the image size went from 107MB to 78.9MB, a reduction in size of ~26%

```bash
❯ docker images
REPOSITORY TAG              IMAGE ID            CREATED             SIZE
py3        latest           95f26289fc0e        4 minutes ago       78.9MB
python     3.8-alpine3.11   6c32e2504283        2 weeks ago         107MB
```

PR #693, opened 2022-02-05
Don't strip debugging symbols in the regular Debian images

### LINKED ISSUE #694: Debian images no longer ship DWARF debugging symbols
Since https://github.com/docker-library/python/pull/687, the regular Debian images no longer ship DWARF symbols, which heavily limits the ability to use gdb and lldb for debugging.

Stripping debugging symbols makes perfect sense to reduce the size of Docker images, but when this optimization was first introduced in https://github.com/docker-library/python/pull/483, the idea was to restrict its use to the slim and Alpine images. That was (unintentionally?) changed in https://github.com/docker-library/python/pull/483, and now debugging symbols are also stripped in the regular Debian images.

WDYT about restoring the previous behavior and limiting this optimization to the slim and Alpine images? (#693).

malor:
Since #687, the regular Debian images no longer ship DWARF symbols, which heavily limits the ability to use gdb and lldb for debugging.

Stripping debugging symbols makes perfect sense to reduce the size of Docker images, but when this optimization was first introduced in #483, the idea was to restrict its use to the slim and Alpine images. That was (unintentionally?) changed in #483, and now debugging symbols are also stripped in the regular Debian images. Undo that change.

Fixes #694 .

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -170,7 +170,9 @@ RUN set -eux; \
 # https://github.com/alpinelinux/aports/commit/2026e1259422d4e0cf92391ca2d3844356c649d0
 		EXTRA_CFLAGS="-DTHREAD_STACK_SIZE=0x100000" \
 {{ ) else "" end -}}
+{{ if is_slim or is_alpine then ( -}}
 		LDFLAGS="-Wl,--strip-all" \
+{{ ) else "" end -}}
 {{ if env.version == "3.7" then ( -}}
 # setting PROFILE_TASK makes "--enable-optimizations" reasonable: https://bugs.python.org/issue36044 / https://github.com/docker-library/python/issues/160#issuecomment-509426916
 		PROFILE_TASK='-m test.regrtest --pgo \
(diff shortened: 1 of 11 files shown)


Dockerfile-linux.template
```diff
@@ -170,7 +170,10 @@ RUN set -eux; \
 # https://github.com/alpinelinux/aports/commit/2026e1259422d4e0cf92391ca2d3844356c649d0
 		EXTRA_CFLAGS="-DTHREAD_STACK_SIZE=0x100000" \
 {{ ) else "" end -}}
+{{ if is_slim or is_alpine then ( -}}
+# strip all symbols to reduce binaries size
```
tianon, line 174: Let's remove this comment -- I think the flag name is pretty clear already, so this isn't necessary:

```suggestion
```

tianon (changes requested):
Oops, nice catch!

Dockerfile-linux.template
malor: Sure! Updated the template and re-ran `apply-template.sh`.

tianon (approved):
Nice, thanks!

PR #697, opened 2022-02-09
Remove coreutils package from Alpine build dependencies

felipecrs:
Because I think that it's not needed.

Refs https://github.com/hassio-addons/addon-base-python/issues/102
Refs https://github.com/hassio-addons/addon-base-python/pull/103

### DIFF
--- a/3.10/alpine3.14/Dockerfile
+++ b/3.10/alpine3.14/Dockerfile
@@ -32,7 +32,6 @@ RUN set -eux; \
 		\
 		bluez-dev \
 		bzip2-dev \
-		coreutils \
 		dpkg-dev dpkg \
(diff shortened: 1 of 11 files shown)


felipecrs:
I suppose the CI can answer this question, so I would appreciate if someone could approve it so that the CI runs.

tianon:
Ah, this is a historical quirk!  It used to be that BusyBox (and thus Alpine) didn't include `nproc`, so we had to install `coreutils` for it (or use `getconf _NPROCESSORS_ONLN`).  At some point that changed, and `busybox` now includes an `nproc` implementation! :smile: :partying_face:

felipecrs:
Awesome! Thank you!

PR #706, opened 2022-03-17
Prevent removing unexpected .a files in case of re-using the Dockerfile to build custom image

belltailjp:
Hello, I'd like to propose a slight change for cleaning up `*.a` which is defined here, so that it doesn't delete unexpected files.
https://github.com/docker-library/python/blob/0047f00c0967161e731c9bab7d50fd95c7c09d46/3.10/slim-buster/Dockerfile#L92

Here is my background:
The current Dockerfile for the dockerhub Python image itself works pretty fine; it doesn't remove anything unexpected, even without my proposal.
However, since this official Dockerfile is so useful, clean and general, we often re-use (borrow) its content to make our own images on top of another image, typically by just rewriting the `FROM` section to something other.
For example, which is my actual case, we build a Python-ready image on top of `nvidia/cuda:11.6.0-xxx`. The issue here is that removing all the `*.a` under `/usr/local` as written in the current Dockerfile also destroys CUDA libraries. This combination is just an example but I think the same thing can happen in other patterns.

Looking at the actual files cleaned up here while building Python 3.10 images for `slim-buster` and `alpine3.15`, I found out `/usr/local/lib/python3.10/config-3.10-x86_64-linux-gnu/libpython3.10.a` is the only `*.a` file removed. So I think modifying the rule for the `find` command from `'*.a'` to `'libpython*.a'` can prevent unexpectedly removing .a files of other libraries in case modifying `FROM` section, while it doesn't change anything on the image for dockerhub.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,23 @@ jobs:
         run: |
           git clone --depth 1 https://github.com/docker-library/bashbrew.git -b master ~/bashbrew
           strategy="$(~/bashbrew/scripts/github-actions/generate.sh)"
+
+          # https://github.com/docker-library/python/pull/706 (ensure we don't have any unexpected ".a" leftovers in "/usr/local")
+          strategy="$(jq <<<"$strategy" -c '
+            .matrix.include |= map(
+              if .os == "ubuntu-latest" then
+                .runs.test += "\n" + (
+                  .meta.entries
+                  | map(
+                    .tags[0]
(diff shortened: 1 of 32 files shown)


tianon:
Interesting!  I'm not _opposed_ but I'd definitely like to do a more thorough check of other versions and variants just to be sure. :sweat_smile: :see_no_evil:

(It'd be even cooler if we could find some way to convince Python not to build or install this file in the first place, but I'm not going to wish too hard for that one :smile:)

belltailjp:
Thanks for the comment!

> more thorough check of other versions and variants just to be sure

Sure! Then maybe I can proceed to run `update.sh` so you and anyone can test it.

tianon:
Nice, that was more successful than I expected. :sweat_smile: :metal:

PR #716, opened 2022-04-09
Remove redundant `--with-system-ffi` configure option

### LINKED ISSUE #715: The `--with-system-ffi` configure option is redundant on Linux for Python 3.7+
Hi!

The Dockerfiles for Linux in this repository currently pass `--with-system-ffi` to configure:
https://github.com/docker-library/python/blob/1999778193a62d95de6816499419be4543f33e67/Dockerfile-linux.template#L162

This option was added back in #205 / #358.

However, as of Python 3.7+ using the system `ffi` library is the default for Linux:
https://bugs.python.org/issue27979
https://github.com/python/cpython/commit/f40d4ddff3c800b3c956a5e8820aabe3aa87cddd#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810

As such, builds are currently emitting this warning during configure:

```
checking for --with-system-ffi... yes
configure: WARNING: --with(out)-system-ffi is ignored on this platform
```

Examples in CI on this repo:
- Python 3.7.13: https://github.com/docker-library/python/runs/5876761054?check_suite_focus=true#step:5:926
- Python 3.11.0a7: https://github.com/docker-library/python/runs/5876759514?check_suite_focus=true#step:5:999

Given that this repo no longer supports Python versions older than Python 3.7, it seems this option can now be dropped?

edmorley:
Since as of Python 3.7+ using system `ffi` is now mandatory (and so the default) on Linux:
https://bugs.python.org/issue27979
https://github.com/python/cpython/commit/f40d4ddff3c800b3c956a5e8820aabe3aa87cddd

Removing the option resolves this configure warning:

```
checking for --with-system-ffi... yes
configure: WARNING: --with(out)-system-ffi is ignored on this platform
```

Closes #715.

### DIFF
--- a/3.10/alpine3.14/Dockerfile
+++ b/3.10/alpine3.14/Dockerfile
@@ -77,7 +77,6 @@ RUN set -eux; \
 		--enable-shared \
 		--with-lto \
 		--with-system-expat \
-		--with-system-ffi \
 		--without-ensurepip \
 	; \
 	nproc="$(nproc)"; \
--- a/3.10/alpine3.15/Dockerfile
+++ b/3.10/alpine3.15/Dockerfile
@@ -77,7 +77,6 @@ RUN set -eux; \
 		--enable-shared \
 		--with-lto \
 		--with-system-expat \
-		--with-system-ffi \
 		--without-ensurepip \
 	; \
 	nproc="$(nproc)"; \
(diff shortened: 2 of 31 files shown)


yosifkit:
Thanks!

PR #732, opened 2022-06-03
Update pip to v22.1

kmecpp:
The current version (22.0.4) doesn't have support for this flag https://github.com/pypa/pip/pull/10990 / https://github.com/pypa/pip/pull/11035

### DIFF
--- a/3.10/alpine3.15/Dockerfile
+++ b/3.10/alpine3.15/Dockerfile
@@ -118,7 +118,7 @@ RUN set -eux; \
 	done
 
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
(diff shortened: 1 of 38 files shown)


tianon:
The intent of the pinning changes we did was to allow Python upstream to determine which version of `pip` we're installing (so that we're consistent with a vanilla install of Python since there are so many opinions on specifically _which_ version of `pip` we should provide).

It does appear that we're finally to a place where the version of `pip` in every supported version of Python is higher than what we were pinning our minimum to (for backwards compatibility with our previous images only), so we should probably remove all the "minimum `pip` version" code from that `versions.sh` file instead.

If you want or need a newer `pip` in your own images, I'd suggest updating it early in your `Dockerfile`.

PR #733, opened 2022-06-03
Remove "minimumPipVersion" code (everything is "new enough" now)

tianon:
Inspired by #732 :heart:


tianon:
(I wanted to remove the `setuptools` bits too, but 3.7 and 3.8 aren't updated to "new enough" setuptools for that to not cause a downgrade yet :sweat_smile:)

PR #774, opened 2022-11-23
Update to Alpine 3.17

zware:
Add Alpine 3.17 and remove 3.15; see https://alpinelinux.org/posts/Alpine-3.17.0-released.html

Hat tip to @jnoordsij's #730, which this PR is modeled after.

### DIFF
--- a/3.10/alpine3.17/Dockerfile
+++ b/3.10/alpine3.17/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM alpine:3.15
+FROM alpine:3.17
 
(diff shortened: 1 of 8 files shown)


matejkloska:
Maybe @tianon might help us and merge? 

We'll soon approach Security SLA for https://nvd.nist.gov/vuln/detail/CVE-2022-42916 and it's fixed only for Alpine 3.17 (curl  7.86.0+)

tianon:
Nice, unlike Ruby and PHP, Python has aggressively backported OpenSSL 3 support to all maintained versions so this just works without any real complications! :smile:

Thanks!

DBendit:
Created https://github.com/docker-library/official-images/pull/13604 to get new images built

PR #783, opened 2023-01-05
Temporarily add an explicit bump from setuptools 65.5.0 to 65.5.1

### LINKED ISSUE #781: Update setuptools version following CVE-2022-40897
Given that the currently bundled version of setuptools (65.5.0) is vulnerable to [this vulnerability](https://github.com/advisories/GHSA-r9hx-vwmv-q579), it would be good to bump that up.

tianon:
Closes #781

### DIFF
--- a/3.10/alpine3.16/Dockerfile
+++ b/3.10/alpine3.16/Dockerfile
@@ -120,7 +120,7 @@ RUN set -eux; \
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
 ENV PYTHON_PIP_VERSION 22.3.1
 # https://github.com/docker-library/python/issues/365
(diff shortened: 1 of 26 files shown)


Quinncuatro:
Excellent! Thanks for the fix!

tianon:
Copying the real important bit from https://github.com/docker-library/python/issues/781#issuecomment-1372524712 to be explicit here:

> if you're using 3.7, 3.8, or 3.9, you'll have no choice but to either update `setuptools` in your own image or try to make a case upstream for them bumping the versions; that being said, I'm guessing they're going to hesitate for the same breaking-changes reason I am, and the likelihood of getting new 57.x or 58.x releases of `setuptools` that fix this CVE with minimal other changes is probably low

Issue #784, opened 2023-01-13
Built Python prefers system packaged libpython when installed; local install should be built with rpath?


ianw:
Hello,

If you happen to pull in the ```libpython``` package from the base distro into a container with the same version of Python, the built python will end up using the libpython shared-library from the system-installed package; e.g.

```console
$ sudo docker run -it --network=host python:3.9-bullseye /bin/bash

# ldd /usr/local/bin/python3.9
	linux-vdso.so.1 (0x0000ffff9cc6a000)
	libpython3.9.so.1.0 => /usr/local/lib/libpython3.9.so.1.0 (0x0000ffff9c89a000)
        ...

# apt-get update
...

# apt-get install -y python3.9-dev
...
The following additional packages will be installed:
  libpython3.9 libpython3.9-dev
The following NEW packages will be installed:
  libpython3.9 libpython3.9-dev python3.9-dev
...

# ldd /usr/local/bin/python3.9
	linux-vdso.so.1 (0x0000ffff82c39000)
	libpython3.9.so.1.0 => /usr/lib/aarch64-linux-gnu/libpython3.9.so.1.0 (0x0000ffff82663000)
        ...
```

AFAICS, ```--enable-shared``` as been there basically forever (https://github.com/docker-library/python/commit/4faf7f1b13a13b24367307acc254bd2d0d48a46e).  From prior comments (https://github.com/docker-library/python/issues/777#issuecomment-1358366197) I believe the intent here is for the ```/usr/local/bin/python``` to be strongly isolated from any packaged python.

One option to avoid this might to be to build twice, a bit like the Debian package does, so that ```/usr/local/bin/python3.X``` is self-contained, and then copy back the ```.so``` library in a second step?

The other option seems to be to add ```rpath``` linker flags to ensure that the ```/usr/local/bin``` version finds its version first.  The binary *could* be patched post-build; I tried that, e.g. (in the same container as above)

```console
# apt-get install patchelf
...

# patchelf --set-rpath '$ORIGIN/../lib' /usr/local/bin/python3.9

# ldd /usr/local/bin/python3.9
	linux-vdso.so.1 (0x0000ffff873dc000)
	libpython3.9.so.1.0 => /usr/local/bin/../lib/libpython3.9.so.1.0 (0x0000ffff86ffd000)
        ...
```

Probably that would take some thought on what other bits need patching like that.

[...]

PR #785, opened 2023-01-13
Fix path to libpython shared-library in python binaries

yosifkit:
- prevent accidental usage of a system installed libpython of the same version

The makes the lookup for `libpython` dependent on where the binary is, so you can do silly things to see that it works:

```console
$ docker run -it --rm 11314a7451b3 sh
/ # ldd /usr/local/bin/python3.11
	/lib/ld-musl-x86_64.so.1 (0x7f9295da7000)
	libpython3.11.so.1.0 => /usr/local/bin/../lib/libpython3.11.so.1.0 (0x7f92957d9000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f9295da7000)
/ # mv /usr/local/ /opt/things/
/ # ldd /opt/things/bin/python3.11
	/lib/ld-musl-x86_64.so.1 (0x7f388530a000)
	libpython3.11.so.1.0 => /opt/things/bin/../lib/libpython3.11.so.1.0 (0x7f3884d3c000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f388530a000)
/ # 
```

Fixes https://github.com/docker-library/python/issues/784


Dockerfile-linux.template
```diff
@@ -173,8 +173,10 @@ RUN set -eux; \
 		EXTRA_CFLAGS="-DTHREAD_STACK_SIZE=0x100000" \
 {{ ) else "" end -}}
 {{ if is_slim or is_alpine then ( -}}
-		LDFLAGS="-Wl,--strip-all" \
-{{ ) else "" end -}}
+		LDFLAGS="-Wl,--strip-all,-rpath='$\$ORIGIN/../lib'" \
+{{ ) else ( -}}
+		LDFLAGS="-Wl,-rpath='$\$ORIGIN/../lib'" \
```
yosifkit, line 178: - `\$` because of the double quotes in the shell to prevent interpolation
- `$$` for `make` to not interpret the `$O`
- `"` because it needed the `'` around the path, but I didn't want `='-Wl,-rpath='"'"'$$ORIGIN/../lib'"'"`

Dockerfile-linux.template
```diff
@@ -173,8 +173,10 @@ RUN set -eux; \
 		EXTRA_CFLAGS="-DTHREAD_STACK_SIZE=0x100000" \
 {{ ) else "" end -}}
 {{ if is_slim or is_alpine then ( -}}
-		LDFLAGS="-Wl,--strip-all" \
-{{ ) else "" end -}}
+		LDFLAGS="-Wl,--strip-all,-rpath='$\$ORIGIN/../lib'" \
```
tianon, line 176:
```suggestion
# \$ because of the double quotes in the shell to prevent interpolation
# $$ for make to not interpret the $O
# " because it needed the ' around the path, but I didn't want ='-Wl,-rpath='"'"'$$ORIGIN/../lib'"'"
		LDFLAGS="-Wl,{{ if is_slim or is_alpine then "--strip-all," else "" end }}-rpath='\$\$ORIGIN/../lib'" \
```

(and then dropping the outer `if`? :eyes:)

tianon (approved):

Issue #787, opened 2023-01-19
Recent updates to 3.9 and 3.7 images led to issue with SSL module during pip deps installation


CarpathianUA:
![image](https://user-images.githubusercontent.com/5362436/213418751-9fd2c1f5-c31e-4563-846d-e7b51bf45db4.png)
```
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError("Can't connect to HTTPS URL because the SSL module is not available.")': /simple/pip/
[1033][2023-01-19 09:56:22] Could not fetch URL https://pypi.org/simple/pip/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='pypi.org', port=443): Max retries exceeded with url: /simple/pip/ (Caused by SSLError("Can't connect to HTTPS URL because the SSL module is not available.")) - skipping
```
`3.9.15` and `3.7.15` images aren't affected

CarpathianUA:
@tianon We use pre-built images  like `python:3.9` as a base layer for our custom imagse

tianon:
Hmm, that's weird -- all that's changed recently is #785, which shouldn't affect TLS, so I wonder what Kaniko is doing that's causing that to misbehave? :thinking:  (unfortunately I don't actually know very much about Kaniko :see_no_evil:)

POD666:
I have no deep understanding on how kaniko works but I will try to explain with my words:

* kaniko is based on `scratch` image by default, so it allows to build any images with any FROM statement.
* after we created custom kaniko image, we had issues building images that are not on the same "base" as our custom kaniko image
* I guess some path changes `-rpath='$\$ORIGIN/../lib'` lead to some conflicts between our custom kaniko image and the new python:3.9 image, but not sure why python:3.9-alpine works fine
* side effect looks like a missing library for SSL. Maybe other libs are missing too.

For anyone else who encounter such issue with custom kaniko - you could use previous python:3.9.15 image.

POD666:
Yes, here is a script to reproduce:
```sh
echo "
FROM gcr.io/kaniko-project/executor:v1.8.1-debug as kaniko
FROM docker:20.10.6
# Add kaniko binary
COPY --from=kaniko /kaniko/executor /kaniko/
# we install a lot of additional tools&packages but the issue is reproduced without them
" > Dockerfile.custom-kaniko

docker build -t custom-kaniko - < Dockerfile.custom-kaniko

# this works with python:3.9.15
[...]

Issue #786, opened 2023-01-18
Latest path changes on 3.10-slim image leading to errors when building applications with staticx


oktavlachs:
Hi everyone and thank you for maintaining this project. I have used your docker images heavily over the past years :)

Since this morning I am facing issues when running staticx against one of my Python applications. I have successfully built an image yesterday and have not changed anything in my application, but with the latest 3.10-slim image, I get an error  with unsupported RPATH/RUNPATH tags, I believe because of the way the paths in the image have changed to absolute paths.
https://github.com/docker-library/python/pull/785

I know this sounds like a very specific issue outside your scope, but I would like to at least know if this is intended on your side or not.
Please find the link to the issue with staticx here: 
https://github.com/JonathonReinhart/staticx/issues/188

Example:
```
...
/tmp/staticx-pyi-1eb5z08n/lib-dynload/_asyncio.cpython-310-x86_64-linux-gnu.so: DT_RUNPATH='/../lib:$ORIGIN/../lib'
...
```

Would be great to get a quick comment on this.

Cheers!

c-gabdalla:
@oktavlachs I found the same errors on official python images, where dtags/runpaths are used even with venvs.

I could circumvent this by using a SO python version, like python 3.8 on debian 11.

The Dockerfile below doesn't work:

```Dockerfile
FROM python:3.11-buster as builder
RUN apt-get update && \
    apt-get install scons patchelf upx -y
WORKDIR /usr/src
COPY requirements.txt .
RUN pip install --no-cache-dir -r requirements.txt
COPY metadata_cleaner.py .
RUN pyinstaller -F metadata_cleaner.py
RUN staticx dist/metadata_cleaner dist/metadata_cleaner_static

FROM debian:10-slim
COPY --from=builder /usr/src/dist/metadata_cleaner_static /usr/local/bin/metadata_cleaner
```

But this one works:

```Dockerfile
FROM debian:11 as builder
RUN apt-get update && \
    apt-get install scons patchelf upx python3-minimal python3-pip python3-venv python3-dev python3-wheel gcc --no-install-recommends -y
ENV VIRTUAL_ENV=/usr/src/.venv 
RUN python3 -m venv $VIRTUAL_ENV
ENV PATH="$VIRTUAL_ENV/bin:$PATH"
WORKDIR /usr/src
COPY requirements.txt .
RUN pip install --no-cache-dir -r requirements.txt
COPY metadata_cleaner.py .
[...]

PR #790, opened 2023-01-20
Better rpath fix

yosifkit:
This reverts https://github.com/docker-library/python/issues/785 and then only changes rpath for python3.x binaries and not the libraries in `/usr/local/lib/python3.x/lib-dynload/` since they use system libraries like libssl (so `$ORIGIN/../lib` doesn't work for them).

Hopefully this should also fix issues like #786 or #787 

Broken, current version:
```console
$ docker run -it --rm python:3.9 bash
root@8f5b65b7cd6b:/# readelf --dynamic /usr/local/lib/python3.9/lib-dynload/_ssl.cpython-39-x86_64-linux-gnu.so

Dynamic section at offset 0x24d58 contains 28 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libssl.so.1.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.1.1]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000001d (RUNPATH)            Library runpath: [/../lib:$ORIGIN/../lib]
...
root@8f5b65b7cd6b:/# # but ldd somehow still works 🤷
root@8f5b65b7cd6b:/# ldd /usr/local/lib/python3.9/lib-dynload/_ssl.cpython-39-x86_64-linux-gnu.so
	linux-vdso.so.1 (0x00007fff459dd000)
	libssl.so.1.1 => /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007fabee010000)
	libcrypto.so.1.1 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.1 (0x00007fabedd1c000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fabedcfa000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fabedb25000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fabedb1f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fabee0da000)
```

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -99,6 +99,7 @@ RUN set -eux; \
 		make \
 		ncurses-dev \
 		openssl-dev \
+		patchelf \
 		pax-utils \
 		readline-dev \
 		sqlite-dev \
@@ -110,10 +111,11 @@ RUN set -eux; \
 		zlib-dev \
 	; \
 	\
-{{ ) elif is_slim then ( -}}
+{{ ) else ( -}}
 	savedAptMark="$(apt-mark showmanual)"; \
 	apt-get update; \
 	apt-get install -y --no-install-recommends \
+{{ if is_slim then ( -}}
 		dpkg-dev \
 		gcc \
 		gnupg dirmngr \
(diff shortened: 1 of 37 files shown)


tianon (approved):

Dockerfile-linux.template
```diff
@@ -224,10 +229,14 @@ RUN set -eux; \
 -}}
 	; \
 	make install; \
+	\
+# https://github.com/docker-library/python/issues/784
+# prevent accidental usage of a system installed libpython of the same version
+	bin="$(readlink -ve /usr/local/bin/python3)"; \
```
tianon, line 235:
```console
+ readlink -ve /usr/local/bin/python3
readlink: unrecognized option: e
BusyBox v1.35.0 (2022-11-19 10:13:10 UTC) multi-call binary.

Usage: readlink [-fnv] FILE

Display the value of a symlink

	-f	Canonicalize by following all symlinks
	-n	Don't add newline
	-v	Verbose
+ bin=
```

Classic BusyBox :smile:

Issue #708, opened 2022-03-24
Include Ubuntu as a variant for faster patch releases


nicktimko:
When using security scanners, there are often complaints about unfixed CVEs in the Debian image. Debian, for better or worse, seems to value increased stability, leading to more, older (found) bugs vs. Ubuntu's quicker releases which probably cause more new bugs. Which is better? _¿Porque no los dos?_

Adding `focal` to the list of variants alongside `buster` and `bullseye` would seem to add Ubuntu to the list of Python images without much trouble. Both `focal` (20.04) and `jammy` (22.04, prerelease until ~Apr 21) tags are available as buildpack-deps tags (https://hub.docker.com/_/buildpack-deps).

I can get a PR together if the maintainers are amenable to the idea.

yosifkit:
My stance remains unchanged: https://github.com/docker-library/python/issues/61#issuecomment-282450114.

nicktimko:
I think the "underlying distribution is irrelevant" doesn't always hold. Frequently, in order to build packages that don't ship `manylinux*` binaries (which don't work on Alpine, so we're past that immediately), some system libraries need to be installed, so you can't really be distro agnostic.

You also can't be agnostic when dealing with security issues. Unless your application is pretty narrow, it's going to be calling out to many other libraries to access a database, format an image, compress a file, send some network traffic...

Because these CVEs come and get squashed, it's hard to "replicate" this issue, but for some example as of now, here's Apache, which has a known vuln in Debian Bullseye, but it's patched in Ubuntu 20.04 Focal. I don't use Apache as much as nginx, but I've had vulnerable libpng's and such show up. *In general* the trend seems to be that Ubuntu security fixes can roll out faster.

* CVE-2022-22720 - Apache OOB vuln
    * [Debian tracker](https://security-tracker.debian.org/tracker/CVE-2022-23943) - `2.4.52-1~deb11u2` is vulnerable
    * [Ubuntu tracker](https://ubuntu.com/security/CVE-2022-23943) patched as of March 14


<details>
<summary>Some commands to check package versions and their output...</summary>

```
$ docker run --pull always --rm -it ubuntu:latest bash -c \
    "apt-get update > /dev/null;apt-cache madison apache2"
latest: Pulling from library/ubuntu
Digest: sha256:bea6d19168bbfd6af8d77c2cc3c572114eb5d113e6f422573c93cb605a0e2ffb
Status: Image is up to date for ubuntu:latest
[...]

PR #791, opened 2023-01-23
Add jammy and kinetic (ubuntu 22.04 and 22.10)

tuler:
My end goal is to build images for `linux/riscv64` architecture.
The problem is that the base used here is [debian](https://hub.docker.com/_/debian) and that doesn't have an official `riscv64` build yet. On the other hand [ubuntu](https://hub.docker.com/_/ubuntu) images already have `riscv64` official builds.

Another separate problem is the integration in the build system, which I believe is based on a Jenkins system, to build riscv64 images. I successfully built riscv64 images in my local arm64 Apple M1 using docker buildx riscv64 support through qemu.

This PR is only about adding ubuntu based images though, which will help to achieve the goal of having riscv64 images.


tianon:
See https://github.com/docker-library/python/issues/708

mudanzasIPROM (commented):
Hola

mudanzasIPROM (commented):
Hjjjj

yosifkit:
Just an FYI, Ubuntu has dropped `riscv64` images, for now: https://github.com/docker-library/official-images/pull/13950#issuecomment-1412111952.

tuler:
Another alternative for adding `linux/riscv64` support for the official python image is to add the `unstable` and `slim-unstable` debian variants.

Those [debian variants](https://hub.docker.com/layers/library/debian/unstable/images/sha256-2482c72e0ad01ae8d942689c725d72429fb579fd37ecf3c61282eb98c2479fb5?context=explore) at docker hub has builds for `linux/riscv64`, as well as [buildpack-deps:unstable](https://hub.docker.com/layers/library/buildpack-deps/unstable/images/sha256-3a887968fe43226ca614e3bdfba591197b3593e08bde65a9a38a179bf5715812?context=explore).

Would you be interested in having a `python:unstable` as well?

tianon:
Unfortunately, as the name implies, Debian Unstable is a fairly unstable base to build on top of, and is frequently broken (by design), so we don't have very many official images based on it (and would prefer not to maintain any ourselves).

tuler:
@cjdcordeiro we are not seeing `linux/riscv64` builds for [buildpack-deps](https://hub.docker.com/layers/library/buildpack-deps/jammy/images/sha256-e902a5ede3335b40edf720db97f0f318f2f014bf34395671c61dd7c554c38967?context=explore) anymore/yet. Will they be added?

cjdcordeiro:
> @cjdcordeiro we are not seeing `linux/riscv64` builds for [buildpack-deps:jammy](https://hub.docker.com/layers/library/buildpack-deps/jammy/images/sha256-e902a5ede3335b40edf720db97f0f318f2f014bf34395671c61dd7c554c38967?context=explore) anymore/yet. Will they be added?

@tuler yes they will. It requires a few changes in our build systems in order to make both i386 and riscv64 coexist across the spectrum of currently supported releases, but we **do intend** do start publishing for riscv64 again. It is currently being worked on but I don't yet have an ETA.

tuler:
> > @cjdcordeiro we are not seeing `linux/riscv64` builds for [buildpack-deps:jammy](https://hub.docker.com/layers/library/buildpack-deps/jammy/images/sha256-e902a5ede3335b40edf720db97f0f318f2f014bf34395671c61dd7c554c38967?context=explore) anymore/yet. Will they be added?
> 
> @tuler yes they will. It requires a few changes in our build systems

@cjdcordeiro any update on this?

tianon:
Sorry, we unfortunately have no plans to support Ubuntu based images here (our matrix is already pretty large supporting all actively supported release lines of Python upstream multiplied by 2x Debian + 2x Alpine + 2x Windows _and_ their respective sets of supported CPU architectures and Python is relatively heavy to build).

cjdcordeiro:
FYI (and especially for @tuler) the riscv64 support for Ubuntu container images is being worked on atm. It is already in QA and should reach the official images soon.

PR #794, opened 2023-01-29
Use the linker to set rpath on just the python binary

jimnicholls:
First build Python. Then re-build just the python executable with the linker flag `-rpath=$ORIGIN/../lib`. With this process, the python binary does not trigger a Rosetta error.

This Dockerfile demonstrates that only the python executable has a runpath set on it. The Python library and other .so files that are built as part of building Python do not have a runpath on them.

<details><summary>Dockerfile</summary>
<p>

```Dockerfile
FROM python:3.10-slim-bullseye

SHELL [ "/bin/bash", "-eux", "-c" ]

RUN \
    apt-get update; \
    apt-get install -y --no-install-recommends binutils patchelf; \
    rm -rf /var/lib/apt/lists/*

RUN \
    /usr/local/bin/python3.10 -V; \
    patchelf --print-rpath /usr/local/bin/python3.10; \
    patchelf --print-rpath /usr/local/lib/libpython3.10.so.1.0; \
    find /usr/local/lib/python3.10 -name '*.so' -print -exec patchelf --print-rpath '{}' \; ; \
    readelf -d /usr/local/bin/python3.10; \
    readelf -d /usr/local/lib/libpython3.10.so.1.0
```

</p>
</details>

<details><summary>docker buildx build --platform=linux/amd64 --progress=plain .</summary>
<p>

```
#6 [3/3] RUN     /usr/local/bin/python3.10 -V;     patchelf --print-rpath /usr/local/bin/python3.10;     patchelf --print-rpath /usr/local/lib/libpython3.10.so.1.0;     find /usr/local/lib/python3.10 -name '*.so' -print -exec patchelf --print-rpath '{}' ; ;     readelf -d /usr/local/bin/python3.10;     readelf -d /usr/local/lib/libpython3.10.so.1.0
#6 0.358 + /usr/local/bin/python3.10 -V
#6 0.374 Python 3.10.9
#6 0.375 + patchelf --print-rpath /usr/local/bin/python3.10
#6 0.389 $ORIGIN/../lib
#6 0.389 + patchelf --print-rpath /usr/local/lib/libpython3.10.so.1.0
#6 0.407
#6 0.409 + find /usr/local/lib/python3.10 -name '*.so' -print -exec patchelf --print-rpath '{}' ';'
#6 0.431 /usr/local/lib/python3.10/lib-dynload/_datetime.cpython-310-x86_64-linux-gnu.so
#6 0.445
[...]

PR #800, opened 2023-02-16
Remove unnecessary make call

avtrifanov:
Seems the second make call appeared due to copy-paste new flags from https://github.com/docker-library/python/issues/784


yosifkit:
It is definitely on purpose to call `make` twice so that we can build just the python binary with the different `LDFAGS` and not apply the `rpath` changes to the libraries in `/usr/local/lib/python*/lib-dynload/`

- #785 
- #790
- #794

PR #817, opened 2023-03-20
Enable buildkit to widen test of its use

tianon:
This image is an especially useful test because it includes both Windows _and_ Linux images (compared to most of the other images that have opted in to buildkit explicitly so far).


tianon:
(This should also be testing https://github.com/docker-library/bashbrew/pull/70 :eyes:)

PR #822, opened 2023-04-03
Add debian bookworm, remove buster

sspans-sbp:
Debian bookworm is in hard-freeze - and will probably be released in the coming weeks.
So it makes sense to have a PR ready to add it once released.

https://lists.debian.org/debian-devel-announce/2023/03/msg00004.html

Feel free to close this if it clashes with other plans, just trying to help.

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -281,7 +281,7 @@ end
 	apt-mark auto '.*' > /dev/null; \
 	apt-mark manual $savedAptMark; \
 	find /usr/local -type f -executable -not \( -name '*tkinter*' \) -exec ldd '{}' ';' \
-		| awk '/=>/ { print $(NF-1) }' \
+		| awk '/=>/ { so = $(NF-1); if (index(so, "/usr/local/") == 1) { next }; gsub("^/(usr/)?", "", so); print so }' \
 		| sort -u \
 		| xargs -r dpkg-query --search \
 		| cut -d: -f1 \
--- a/3.10/bookworm/Dockerfile
+++ b/3.10/bookworm/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM buildpack-deps:buster
+FROM buildpack-deps:bookworm
 
 # ensure local python is preferred over distribution python
 ENV PATH /usr/local/bin:$PATH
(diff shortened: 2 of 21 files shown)


tianon:
As in previous cycles, this will also mean we remove support for Debian Buster (as the compilation of Python is very heavy, so we stick to only maintaining builds for the latest two versions of each supported distribution).

sspans-sbp:
Testing pipeline showed that sqlite3 and gdbm were purged from the slim images.
Which breaks the related python modules. I've opted to install the packages manually to make sure they're not purged.

Alternatively the ldd/find could be made smarter somehow.

sspans:
FYI: bookwork release date has been announced as 2023-06-10.

https://lists.debian.org/debian-devel-announce/2023/04/msg00007.html

aqeelat:
@sspans-sbp Debian 12 is finally out now. Could we get this PR merged?

sspans-sbp:
I've re-generated the PR now that bookworm has been released.

It looks good to me, but I'll let @tianon judge the usrmerge change

Dockerfile-linux.template
```diff
@@ -283,6 +283,8 @@ end
 	find /usr/local -type f -executable -not \( -name '*tkinter*' \) -exec ldd '{}' ';' \
 		| awk '/=>/ { print $(NF-1) }' \
 		| sort -u \
+		| awk '{ print $1; system("realpath " $1) }' \
+		| sort -u \
```
tianon, line 287: Some light "awk(ward) golf" :sweat_smile:

This essentially takes paths like `/lib/foo.so` and `/usr/lib/foo.so` and normalizes _both_ to `lib/foo.so`, which is more likely to match accurately in `dpkg-query --search` no matter which path the `.deb` happened to record.  I also figured it's a good opportunity to make sure we ignore `/usr/local` entries :smile:

```suggestion
		| awk '/=>/ { so = $(NF-1); if (index(so, "/usr/local/") == 1) { next }; gsub("^/(usr/)?", "", so); print so }' \
		| sort -u \
```

(I'm happy to take over from here if you'd prefer, but I want to give you first crack since you've done all the work so far :+1:)

sspans-sbp:
@tianon great suggestion, merged and regenerated.

tianon:
(while reviewing, I noticed this missed `versions.sh` updates, so while updating those I also squashed :heart:)

aqeelat:
@tianon Thank you for merging the PR. How long does it take for the images to be published? The dockerhub repo was updated 11 hours ago and still does not include bookworm.

sspans-sbp:
@aqeelat The [official-images](https://github.com/docker-library/official-images) repository has not been updated yet.

I'm not quite sure if that's an automated or manual procedure. 
But @tianon is listed as a maintainer - so he will get to it in due time.

tianon:
:eyes: https://github.com/docker-library/official-images/pull/14854

PR #830, opened 2023-05-16
Enforce LF line-endings for all files

orgads:
* Required for getting the shebang right (for example, in WSL).
* Fixes jq failures on CRLF Dockerfiles:

```
jq: error: syntax error, unexpected INVALID_CHARACTER (Unix shell quoting issues?) at <top-level>, line 1:

jq: 1 compile error
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 /*/**/Dockerfile       linguist-generated
 /Dockerfile*.template  linguist-language=Dockerfile
+*                      text=auto eol=lf


.gitattributes
```diff
@@ -1,2 +1,4 @@
-/*/**/Dockerfile       linguist-generated
-/Dockerfile*.template  linguist-language=Dockerfile
+/*/**/Dockerfile       linguist-generated eol=lf
+/Dockerfile*.template  linguist-language=Dockerfile eol=lf
+*.sh                   eol=lf
+*.awk                  eol=lf
```
tianon, line 4: I was hoping we could avoid being quite this explicit, but this is fair -- I think we can simplify by generalizing to _all_ text files, though:

```suggestion
*                      text=auto eol=lf
/*/**/Dockerfile       linguist-generated
/Dockerfile*.template  linguist-language=Dockerfile
```

(If I'm reading https://git-scm.com/docs/gitattributes correctly)

orgads: Thanks, done.

PR #831, opened 2023-05-16
Disable pip cache

orgads:
It is useless in docker envs, and just takes space.

### DIFF
--- a/3.10/alpine3.17/Dockerfile
+++ b/3.10/alpine3.17/Dockerfile
@@ -136,6 +136,7 @@ ENV PYTHON_SETUPTOOLS_VERSION 65.5.1
 # https://github.com/pypa/get-pip
 ENV PYTHON_GET_PIP_URL https://github.com/pypa/get-pip/raw/0d8570dc44796f4369b652222cf176b3db6ac70e/public/get-pip.py
 ENV PYTHON_GET_PIP_SHA256 96461deced5c2a487ddc65207ec5a9cffeca0d34e7af7ea1afc470ff0d746207
+ENV PIP_NO_CACHE_DIR 1
 
 RUN set -eux; \
 	\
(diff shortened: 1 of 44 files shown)


ad-m-ss (commented):
LGTM 👍 

FTR: Downstream image may leverage pip cache by:

```
FROM python:3.10.7-bullseye AS build-pip
ENV PIP_CACHE_DIR=/var/cache/buildkit/pip
RUN mkdir -p $PIP_CACHE_DIR
RUN --mount=type=cache,target=/var/cache/buildkit/pip \
    pip install -U pip wheel
```

orgads:
> FTR: Downstream image may leverage pip cache by:
> 
> ```
> FROM python:3.10.7-bullseye AS build-pip
> ENV PIP_CACHE_DIR=/var/cache/buildkit/pip
> RUN mkdir -p $PIP_CACHE_DIR
> RUN --mount=type=cache,target=/var/cache/buildkit/pip \
>     pip install -U pip wheel
> ```

Not exactly. You'll also need:
```
ENV PIP_NO_CACHE_DIR=
```

tianon:
I definitely sympathize with the motivation here (and want this myself for my own `FROM python` images!), but I do worry about the impact of setting explicit defaults for these, _especially_ when disabling our tweaks to the defaults is (as you've demonstrated) not entirely straightforward/obvious.  Even if this is what most users actually want, it _is_ a surprising moment when they discover that the defaults they thought applied (and that the `pip` documentation describes applying!) actually do _not_ seem to apply in practice. :sweat_smile:

orgads:
I understand your concern. I'll try to check what other popular image do wrt caching.

orgads:
Ok, you're probably right. I also see that it is a common practice to use a mount for the cache, which will be broken if this change is accepted.

So maybe we should just recommend setting this variable in the README instead?

orgads:
Actually I see that the [documentation](https://hub.docker.com/_/python/) of the image already uses `--no-cache-dir` in the example, so I guess no further action is needed:

```
RUN pip install --no-cache-dir -r requirements.txt
```
(notice that this affects only the packages that are installed during the build, and has no effect when running pip in the container, but this is less important).

Also this doesn't cover packages that require build. See https://github.com/pypa/pip/issues/12031

PR #834, opened 2023-05-26
Use latest pip & setuptools for python 3.9

pgrz:
What's the reason for not using latest pip & setuptools?

Disclaimer: This PR only covers python 3.9 as I'm personally interested in this one only :)


ad-m-ss:
See #781 . We pin to the version of setuptools that comes bundled with each Python release.

pgrz:
Got it, thanks

PR #851, opened 2023-07-17
Move "get-pip" URL to a separate versions.json section so commits are more clear

tianon:
This should help make commits like these more obvious at a glance (updating nothing but "get-pip"):

- https://github.com/docker-library/python/commit/4d12931: Update 3.9
- https://github.com/docker-library/python/commit/aaced40: Update 3.8
- https://github.com/docker-library/python/commit/597175b: Update 3.7
- https://github.com/docker-library/python/commit/b43fc27: Update 3.12-rc
- https://github.com/docker-library/python/commit/1af036f: Update 3.11
- https://github.com/docker-library/python/commit/00dce75: Update 3.10


versions.sh
```diff
@@ -179,6 +179,9 @@ for version in "${versions[@]}"; do
 			version: env.fullVersion,
 			pip: {
 				version: env.pipVersion,
+			},
+			"get-pip": {
+				version: env.getPipCommit,
```
tianon, line 184: Maybe this should be a full URL like https://github.com/pypa/get-pip/commit/12b30d8c229961b6c0892101b365ec185fab7e14 so GitHub shortens them to be prettier?

yosifkit: Yeah, full URL seems nicer.

tianon: Updated!

PR #883, opened 2023-11-02
add symlinks in /usr/bin

anarcat:
Using the stock python:slim container is really difficult if you already have an existing script with `#!/usr/bin/python3` on top. You will get a confusing error like:

```
/usr/bin/bash: line 140: ./check.py: cannot execute: required file not found
```

... while the script otherwise works perfectly well. I understand that it's relatively standard to install things in `/usr/local` and I generally support this, but in this very specific case, it seems extremely backwards.

I also understand that, normally, you should *install* a script with an `entrypoint` and so on, but we shouldn't let perfection be the enemy of good here.

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -256,6 +256,7 @@ RUN set -eux; \
 		[ -s "/usr/local/bin/$src" ]; \
 		[ ! -e "/usr/local/bin/$dst" ]; \
 		ln -svT "$src" "/usr/local/bin/$dst"; \
+		ln -svT "/usr/local/bin/$src" "/usr/bin/$dst"; \
 	done
 
 # if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"


yosifkit:
Unfortunately, this would conflict with Debian's `python3` package that can easily come in as a dependency of another installed package (and would then be the wrong `python3` for that package).

anarcat:
On 2023-11-02 13:05:55, yosifkit wrote:
> Unfortunately, this would conflict with Debian's `python3` package that can easily come in as a dependency of another installed package (and would then be the wrong `python3` for that package).

Would it?

I actually wonder: the package would probably overwrite the symlink,
which would be fine, no?

LaurentGoderre:
Maybe Im missing something but wouldn't it ber simper to modify the shebang to use the syntax often used for referencing bash

```sh
/usr/bin/env python3

your script
```

anarcat:
On 2023-11-03 07:58:47, Laurent Goderre wrote:
> Maybe Im missing something but wouldn't it ber simper to modify the shebang to use the syntax often used for referencing bash
>
> ```sh
> /usr/bin/env python3
>
> your script
> ```

That means changing absolutely every random script out there, definitely
not simpler in my case.

tianon:
Generally yes, installing `python3` would overwrite the symlink and be fine, but the part that makes this complicated is that for many of our variants it's already pre-installed (because it's a pre-requisite of many of the common tools included in our "batteries included" images), so there's not a specific "install" that would fix it - we'd just have pre-installed tools that may or may not have different behavior (and some that might not work at all).

yosifkit:
Closing since this is not something we can reasonably add to the images (see @tianon's [comment](https://github.com/docker-library/python/pull/883#issuecomment-1834698677)).

Users can add the symlinks in their own dependent image or run their python script without relying on the shebang, i.e. via `python3 ./script.py`.

PR #912, opened 2024-03-20
Update to 3.10.14 manually for now

tianon:
This is necessary to account for an upstream release-signing snafu where the 3.8 and 3.9 release manager also signed the latest 3.10 release. 😄

See https://github.com/python/cpython/issues/117053 for details (and to track the resolution).


tianon:
Since this is a security release, we don't want to wait for a reply to https://github.com/python/cpython/issues/117053#issuecomment-2009768879 in order to get the builds out, especially since it's not like it's signed by a random _rogue_ key (Łukasz is the 3.8 and 3.9 release manager _and_ wrote the blog post about the security releases :smile:).

We can easily fix this again later if/when there's a resolution. :+1:

yosifkit (approved):

PR #931, opened 2024-06-13
Attempt to speed up riscv64 builds

yosifkit:
Some Python riscv64 images currently fail to complete building and hit the 3-hour Jenkins job limit. 

Currently that is `python:3.13.0b2-alpine3.20`, `python:3.12.4-alpine3.20`, `python:3.11.9-alpine3.20`, `python:3.10.14-alpine3.20`

This change is an attempt to get them to build in the allotted time instead of just dropping `riscv64` on these Python versions. A less optimized python is better than no python image at all.

Build speed when emulating `riscv64` from `amd64` WSL (~14 cores, so a bit more than our `riscv64` builder):
| python  | with optimizations |    without |
|---------|-------------------:|-----------:|
| 3.9     |         15 minutes |         \- |
| 3.10    |         52 minutes | 23 minutes |
| 3.11    |         92 minutes | 47 minutes |
| 3.12    |        103 minutes | 50 minutes |
| 3.13-rc |        109 minutes |    not tested|

### DIFF
--- a/3.10/alpine3.19/Dockerfile
+++ b/3.10/alpine3.19/Dockerfile
@@ -73,7 +73,7 @@ RUN set -eux; \
 	./configure \
 		--build="$gnuArch" \
 		--enable-loadable-sqlite-extensions \
-		--enable-optimizations \
+		$(test "$gnuArch" != 'riscv64-linux-musl' && echo '--enable-optimizations') \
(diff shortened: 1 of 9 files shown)


tianon (approved):
Sure, let's give it a go :rocket:

PR #935, opened 2024-07-10
Remove riscv64 from python versions that are too slow to build

yosifkit:
https://github.com/docker-library/python/pull/931 was enough for Python `3.10` 🎉, but not `3.11`, `3.12`, or `3.13-rc` 😞😢. They still take more than 3 hours to build. 😱

Not sure if there is a way to bring them back. 🤔 `riscv64` is not a supported platform (https://peps.python.org/pep-0011/) and I doubt the compile time of Python itself is a goal of the project (or of `gcc`).

image

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -137,6 +137,10 @@ for version; do
 		case "$version" in
 			3.8 | 3.9) ;;
 			*)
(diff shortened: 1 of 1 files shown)


tianon (approved):
:sob: :heart:

PR #954, opened 2024-08-23
Stop installing setuptools in python 3.12+

### LINKED ISSUE #952: Stop including setuptools and wheel in Python 3.12+ images
In Python 3.12, the stdlib's `ensurepip` and `venv` modules were updated to no longer install `setuptools` alongside pip. This was viable since as of pip v22.1, for pre-PEP-517/518 packages [pip will now](https://github.com/pypa/pip/pull/10717) default to the isolated build environment mode (along with a fallback legacy setuptools build backend, with setuptools and wheel automatically installed), if the setuptools package isn't installed globally.

See:
https://github.com/python/cpython/issues/95299
https://github.com/python/cpython/commit/ece20dba120a1a4745721c49f8d7389d4b1ee2a7

As such, when this repo added support for Python 3.12, the explicit setuptools requirement was removed, as part of:
https://github.com/docker-library/python/pull/833

However, that change didn't actually remove setuptools from the Python 3.12+ images, since get-pip [implicitly installs setuptools and wheel](https://github.com/pypa/get-pip/blob/ac9253767f0be4cdef620a3ca07187a8cabc66a5/templates/default.py#L46-L78) if they are not already installed. Instead, the result was only that the setuptools version is now unconstrained and get-pip will pull in whatever is the latest version of setuptools at the time of the image being built.

eg:

```
$ docker run -q --rm python:3.12 pip list
Package    Version
---------- -------
pip        24.2
setuptools 72.2.0
wheel      0.44.0
```

This means that the Python 3.12 (and 3.13-rc) images here:
(a) are less aligned with the experience one gets when using `ensurepip` or `venv` from the stdlib (which no longer install setuptools),
(b) now have a floating setuptools version when they didn't before (which doesn't seem ideal from a determinism point of view, particularly given intentional breaking changes in recent setuptools major releases as part of an ongoing tech-debt clean-up effort).

I think it would be best if the packages in these images were consistent with the environment created by `ensurepip` / `venv`. Which for Python 3.12+ would mean not shipping with setuptools or wheel.

There are a few ways this could be done:
1. For Python 3.12+, pass `--no-setuptools` and `--no-wheel` when invoking get-pip.py.
2. By switching from get-pip to ensurepip (#951), which would (a) mean these images automatically match the behaviour of the stdlib, (b) mean a lot of the version pinning/handling code can be simplified (since ensurepip defaults to installing the bundled version, not the latest version from PyPI).
3. By waiting for when https://github.com/pypa/get-pip/pull/218 is merged. Though at that point, there is even less reason for this repo to keep using get-pip instead of ensurepip, so we'd probably want to do (2) anyway.

See also:
- https://github.com/pypa/virtualenv/issues/2487
- https://github.com/pypa/get-pip/pull/218

yosifkit:
This fixes the intent that was meant to happen in #833 of no longer including setuptools (or wheel).

Fixes #952.

If you are using Python 3.12+ and you have build or runtime dependencies on setuptools or wheel, a simple addition to your Dockerfile is all that is needed. (I think longer term, they might make sense as additions to your `requirements.txt`)
```dockerfile
FROM python:3.12
WORKDIR /usr/src/app

### ⬇️ this line ⬇️ ##########################
RUN pip install --no-cache-dir setuptools wheel
###############################################

# continue your regular installation
COPY requirements.txt ./
RUN pip install --no-cache-dir -r requirements.txt

# ...
```

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -307,7 +307,12 @@ RUN set -eux; \
 		"pip==$PYTHON_PIP_VERSION" \
 {{ if .setuptools then ( -}}
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
-{{ ) else "" end -}}
+		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		wheel \
+{{ ) else ( -}}
+		--no-setuptools \
+		--no-wheel \
+{{ ) end -}}
 	; \
 	rm -f get-pip.py; \
 	\
--- a/3.10/alpine3.19/Dockerfile
+++ b/3.10/alpine3.19/Dockerfile
@@ -151,6 +151,8 @@ RUN set -eux; \
 		--no-compile \
 		"pip==$PYTHON_PIP_VERSION" \
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		wheel \
 	; \
 	rm -f get-pip.py; \
 	\
--- a/3.10/alpine3.20/Dockerfile
+++ b/3.10/alpine3.20/Dockerfile
@@ -151,6 +151,8 @@ RUN set -eux; \
 		--no-compile \
 		"pip==$PYTHON_PIP_VERSION" \
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		wheel \
 	; \
 	rm -f get-pip.py; \
 	\
--- a/3.10/bookworm/Dockerfile
+++ b/3.10/bookworm/Dockerfile
@@ -121,6 +121,8 @@ RUN set -eux; \
 		--no-compile \
 		"pip==$PYTHON_PIP_VERSION" \
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		wheel \
 	; \
 	rm -f get-pip.py; \
 	\
--- a/3.10/bullseye/Dockerfile
+++ b/3.10/bullseye/Dockerfile
@@ -121,6 +121,8 @@ RUN set -eux; \
 		--no-compile \
 		"pip==$PYTHON_PIP_VERSION" \
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		wheel \
 	; \
 	rm -f get-pip.py; \
 	\
--- a/3.10/slim-bookworm/Dockerfile
+++ b/3.10/slim-bookworm/Dockerfile
@@ -164,6 +164,8 @@ RUN set -eux; \
 		--no-compile \
 		"pip==$PYTHON_PIP_VERSION" \
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		wheel \
 	; \
 	rm -f get-pip.py; \
 	\
--- a/3.10/slim-bullseye/Dockerfile
+++ b/3.10/slim-bullseye/Dockerfile
@@ -164,6 +164,8 @@ RUN set -eux; \
 		--no-compile \
 		"pip==$PYTHON_PIP_VERSION" \
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
+		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		wheel \
 	; \
 	rm -f get-pip.py; \
 	\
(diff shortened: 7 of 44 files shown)


Dockerfile-linux.template
```diff
@@ -307,7 +307,10 @@ RUN set -eux; \
 		"pip==$PYTHON_PIP_VERSION" \
 {{ if .setuptools then ( -}}
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
-{{ ) else "" end -}}
+{{ ) else ( -}}
+		--no-setup-tools \
```
tianon, line 311:
```suggestion
		--no-setuptools \
```

https://github.com/pypa/get-pip/blob/ac9253767f0be4cdef620a3ca07187a8cabc66a5/README.md#get-pippy-options :eyes:

Dockerfile-linux.template
```diff
@@ -307,10 +307,23 @@ RUN set -eux; \
...
 	\
-	pip --version
+	pip --version; \
+{{ if .setuptools then ( -}}
+	# smoke test to ensure that setuptools/wheel are included
+	python -c 'import setuptools'; \
+	python -c 'import wheel'
+{{ ) else ( -}}
+	# smoke test to ensure that setuptools/wheel are not included
+	if python -c 'import setuptools' || python -c 'import wheel'; then \
```
yosifkit, line 324:
```console
+ python -c import setuptools
ModuleNotFoundError: No module named 'setuptools'
+ python -c import wheel
ModuleNotFoundError: No module named 'wheel'
```
🤔 Maybe these could be `2>/dev/null`

Dockerfile-linux.template
```diff
@@ -307,10 +307,23 @@ RUN set -eux; \
...
+		--no-setuptools \
+		--no-wheel \
+{{ ) end -}}
 	; \
 	rm -f get-pip.py; \
 	\
-	pip --version
+	pip --version; \
+{{ if .setuptools then ( -}}
+	# smoke test to ensure that setuptools/wheel are included
```
edmorley, line 319: Longer term should these smoke tests live in `docker-library/official-images` where the other tests are? eg:
https://github.com/docker-library/official-images/tree/master/test/tests/python-imports
https://github.com/docker-library/official-images/tree/master/test/tests/python-no-pyc

Dockerfile-linux.template
```diff
@@ -307,10 +307,23 @@ RUN set -eux; \
 		"pip==$PYTHON_PIP_VERSION" \
 {{ if .setuptools then ( -}}
 		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
```
edmorley, line 309: For completeness, and in case `get-pip` were to ever stop installing `wheel` by default for older Python versions, should we add `wheel` as an explicit dependency here (without a version specfier, to match the `get-pip` behaviour)?

ie:
```suggestion
		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
		wheel \
```

Dockerfile-linux.template
edmorley, line 1: I think the Windows template will need similar changes too:
https://github.com/docker-library/python/blob/cc2cf19f2c9a440706d17b5937bf104052eef967/Dockerfile-windows.template#L74-L81

edmorley (commented):
Thank you for opening this! :-)

edmorley:
I think it might also be worth updating the PR description to give an example of how to restore the previous behaviour, for the small subset of users who are running broken dependencies (that don't correctly list an explicit dependency on setuptools/wheel) or are running pip with build isolation disabled.

For example:

```Dockerfile
FROM python:3

WORKDIR /usr/src/app

RUN pip install --no-cache-dir setuptools wheel
COPY requirements.txt ./
RUN pip install --no-cache-dir -r requirements.txt

# ...
```

(this is an adapted version of the `Dockerfile` usage example listed at https://hub.docker.com/_/python/)

Dockerfile-linux.template
tianon: I guess relevant to https://github.com/docker-library/python/pull/954#discussion_r1729987390, if we made these a dedicated script we could probably do a single script that's more complex, like catching the import exception directly in Python code instead?

Dockerfile-linux.template
tianon: Ah yeah, that's probably a good idea, even though I don't love it :joy: :sob: :heart:

Dockerfile-linux.template
yosifkit: Yeah, that makes sense. I removed them so they can just be in the test scripts.

tianon:
https://github.com/docker-library/official-images/pull/17446 :eyes: (I've verified that works with this PR's images, and fails on the current images)

tianon:
(Close and reopen to force trigger _new_ CI so it includes our updated test. :bow:  I've also got https://github.com/docker-library/python/actions/runs/10623880671 running fresh on master so we can see the failing baseline.)

tianon:
Hilariously, thanks to https://github.com/pypa/get-pip/pull/218, my "baseline" test on master is succeeding because now `get-pip.py` has the same behavior as this PR. :joy:

tianon:
Now we're being really explicit about it both directions, which feels right. :smile:

PR #955, opened 2024-08-30
Switch from get-pip.py to ensurepip

### LINKED ISSUE #951: Switch from get-pip to ensurepip
The images in this repo currently use https://github.com/pypa/get-pip to bootstrap pip, however, all versions of Python supported here now include the `ensurepip` module (which was added in Python 3.4), which could be used instead (along with a suitable `pip install -U` command to upgrade to the chosen pinned pip etc):
https://docs.python.org/3/library/ensurepip.html

Switching to `ensurepip` would remove an external dependency and the script version bumping that comes with it (eg 1980d7b46875e86518a0fb1fff9da4c39f5261ea x8 for each upstream get-pip change).

Would you be open to a PR that made that switch?

edmorley:
Since:
* All versions of Python that are actively built by this repo now include the `ensurepip` module.
* The policy of these images is now to use the same pip version as the one bundled with `ensurepip` (rather than always upgrading as pip releases occur) to avoid breaking changes, and for parity with the `venv` module.
* As such, we might as well actually use `ensurepip` to install pip (since it installs the exact pip version we want) rather than manually doing the same using `get-pip.py`.
* Doing so simplifies the build scripts and removes an external dependency, tightening the supply chain.

Now that the pip/setuptools versions track (or mostly track, in the case of setuptools) the ensurepip versions, the concerns over frequent invalidation of the Python layer no longer apply, and so the pip/setuptools install can now be part of the Python layer, reducing layer count by one.

This change is a no-op in terms of pip/setuptools/wheel versions, since the pip versions being used already exactly matched the `ensurepip` version of pip.

Closes #951.

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -176,7 +176,7 @@ RUN set -eux; \
 		--with-lto \
 {{ ) end -}}
 		--with-system-expat \
-		--without-ensurepip \
+		--with-ensurepip \
 	; \
 	nproc="$(nproc)"; \
 {{ if is_alpine then ( -}}
@@ -259,63 +259,27 @@ RUN set -eux; \
 {{ ) else "" end -}}
 {{ ) end -}}
 	\
-	python3 --version
+	export PYTHONDONTWRITEBYTECODE=1; \
+	python3 --version; \
+{{ if .setuptools then ( -}}
+	\
+	pip3 install \
+		--disable-pip-version-check \
+		--no-cache-dir \
+		--no-compile \
+		{{ "setuptools==\( .setuptools.version )" | @sh }} \
+		wheel \
+	; \
+{{ ) else "" end -}}
+	pip3 --version
 
 # make some useful symlinks that are expected to exist ("/usr/local/bin/python" and friends)
 RUN set -eux; \
-	for src in idle3 pydoc3 python3 python3-config; do \
+	for src in idle3 pip3 pydoc3 python3 python3-config; do \
 		dst="$(echo "$src" | tr -d 3)"; \
 		[ -s "/usr/local/bin/$src" ]; \
 		[ ! -e "/usr/local/bin/$dst" ]; \
 		ln -svT "$src" "/usr/local/bin/$dst"; \
 	done
-
-# if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
-ENV PYTHON_PIP_VERSION {{ .pip.version }}
-{{ if .setuptools then ( -}}
-# https://github.com/docker-library/python/issues/365
-ENV PYTHON_SETUPTOOLS_VERSION {{ .setuptools.version }}
-{{ ) else "" end -}}
-# https://github.com/pypa/get-pip
-ENV PYTHON_GET_PIP_URL {{ ."get-pip".url }}
-ENV PYTHON_GET_PIP_SHA256 {{ ."get-pip".sha256 }}
-
-RUN set -eux; \
-	\
-{{ if is_slim then ( -}}
-	savedAptMark="$(apt-mark showmanual)"; \
-	apt-get update; \
-	apt-get install -y --no-install-recommends wget; \
-	\
-{{ ) else "" end -}}
-	wget -O get-pip.py "$PYTHON_GET_PIP_URL"; \
-	echo "$PYTHON_GET_PIP_SHA256 *get-pip.py" | sha256sum -c -; \
-	\
-{{ if is_slim then ( -}}
-	apt-mark auto '.*' > /dev/null; \
-	[ -z "$savedAptMark" ] || apt-mark manual $savedAptMark > /dev/null; \
-	apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false; \
-	rm -rf /var/lib/apt/lists/*; \
-	\
-{{ ) else "" end -}}
-	export PYTHONDONTWRITEBYTECODE=1; \
-	\
-	python get-pip.py \
-		--disable-pip-version-check \
-		--no-cache-dir \
-		--no-compile \
-		"pip==$PYTHON_PIP_VERSION" \
-{{ if .setuptools then ( -}}
-		"setuptools==$PYTHON_SETUPTOOLS_VERSION" \
-		# get-pip.py installs wheel by default, adding in case get-pip defaults change
-		wheel \
-{{ ) else ( -}}
-		--no-setuptools \
-		--no-wheel \
-{{ ) end -}}
-	; \
-	rm -f get-pip.py; \
-	\
-	pip --version
 
 CMD ["python3"]
--- a/Dockerfile-windows.template
+++ b/Dockerfile-windows.template
@@ -22,7 +22,7 @@ RUN $url = ('https://www.python.org/ftp/python/{0}/python-{1}-amd64.exe' -f ($en
 			'PrependPath=1', \
 			'Shortcuts=0', \
 			'Include_doc=0', \
-			'Include_pip=0', \
+			'Include_pip=1', \
 			'Include_test=0' \
 		) \
 	).ExitCode; \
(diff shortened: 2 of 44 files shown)


Dockerfile-linux.template
```diff
@@ -259,63 +259,26 @@ RUN set -eux; \
...
+	export PYTHONDONTWRITEBYTECODE=1; \
+	python3 --version; \
+{{ if .setuptools then ( -}}
+	\
+	pip3 install \
+		--disable-pip-version-check \
+		--no-cache-dir \
+		--no-compile \
+		"setuptools=={{ .setuptools.version }}" \
+		wheel; \
```
tianon, line 271:
```suggestion
		wheel \
	; \
```

Dockerfile-windows.template
```diff
@@ -42,45 +42,18 @@ RUN $url = ('https://www.python.org/ftp/python/{0}/python-{1}-amd64.exe' -f ($en
...
+		'setuptools=={{ .setuptools.version }}' \
 		# get-pip.py installs wheel by default, adding in case get-pip defaults change
-		wheel \
-{{ ) else ( -}}
-		--no-setuptools \
-		--no-wheel \
-{{ ) end -}}
-	; \
-	Remove-Item get-pip.py -Force; \
+		wheel; \
```
tianon, line 55:
```suggestion
		wheel \
	; \
```

Dockerfile-linux.template
```diff
@@ -259,63 +259,26 @@ RUN set -eux; \
...
-	python3 --version
+	export PYTHONDONTWRITEBYTECODE=1; \
+	python3 --version; \
+{{ if .setuptools then ( -}}
+	\
+	pip3 install \
+		--disable-pip-version-check \
+		--no-cache-dir \
+		--no-compile \
+		"setuptools=={{ .setuptools.version }}" \
```
tianon, line 270:
```suggestion
		{{ "setuptools==\( .setuptools.version )" | @sh }} \
```

Dockerfile-windows.template
```diff
@@ -42,45 +42,18 @@ RUN $url = ('https://www.python.org/ftp/python/{0}/python-{1}-amd64.exe' -f ($en
...
+	Write-Host ('Installing setuptools and wheel ...'); \
+	pip install \
 		--disable-pip-version-check \
 		--no-cache-dir \
 		--no-compile \
-		('pip=={0}' -f $env:PYTHON_PIP_VERSION) \
-{{ if .setuptools then ( -}}
-		('setuptools=={0}' -f $env:PYTHON_SETUPTOOLS_VERSION) \
+		'setuptools=={{ .setuptools.version }}' \
 		# get-pip.py installs wheel by default, adding in case get-pip defaults change
```
tianon, line 54:
```suggestion
```

versions.sh
```diff
@@ -171,22 +159,12 @@ for version in "${versions[@]}"; do
 			;;
 	esac
 
-	# TODO wheelVersion, somehow: https://github.com/docker-library/python/issues/365#issuecomment-914669320
-
-	echo "$version: $fullVersion (pip $pipVersion${setuptoolsVersion:+, setuptools $setuptoolsVersion}${hasWindows:+, windows})"
+	echo "$version: $fullVersion (${setuptoolsVersion:+setuptools $setuptoolsVersion}${hasWindows:+${setuptoolsVersion:+, }windows})"
```
tianon, line 162: Hmm, this means 3.12+ will output `3.12: 3.12.XX ()` eventually once it's at the point where the Windows builds drop, but I think that's probably OK (I don't think there's a significantly cleaner way to write this that isn't going to be overkill just for this one edge case that's going to be a while before we hit :smile:).

tianon (commented):
Excellent, thank you!  Just a few really minor stylistic notes, but overall this looks really good. :heart:

versions.sh
edmorley: Ah good point :-)

What do you think about this?

```bash
	if [[ -n "${setuptoolsVersion}" || -n "${hasWindows}" ]]; then
		echo "$version: $fullVersion (${setuptoolsVersion:+setuptools $setuptoolsVersion}${hasWindows:+${setuptoolsVersion:+, }windows})"
	else
		echo "$version: $fullVersion"
	fi
```

Dockerfile-windows.template
```diff
@@ -42,45 +42,18 @@ RUN $url = ('https://www.python.org/ftp/python/{0}/python-{1}-amd64.exe' -f ($en
...
-	python get-pip.py \
+	Write-Host ('Installing setuptools and wheel ...'); \
+	pip install \
 		--disable-pip-version-check \
 		--no-cache-dir \
 		--no-compile \
-		('pip=={0}' -f $env:PYTHON_PIP_VERSION) \
-{{ if .setuptools then ( -}}
-		('setuptools=={0}' -f $env:PYTHON_SETUPTOOLS_VERSION) \
+		'setuptools=={{ .setuptools.version }}' \
```
edmorley, line 53: @tianon Would you like the same change made to this line (in the Windows template) as you suggested to the linux template here?
https://github.com/docker-library/python/pull/955/files#r1739349502

(I'm not familiar with the templating language, so am unclear what the `| @sh` modifier does)

versions.sh
tianon: I'm thinking maybe the other extreme and just drop the extra data -- the `setuptools` versions won't be updated in a way that isn't tied to the Python version, and the `windows` is just a debugging tool to make sure the Windows versions get picked up correctly which is honestly something we'll only notice because they drop unexpectedly at which point we'll be explicitly debugging _that_ and probably using `bash -x` to do so, so it's really not providing much value anymore. :thinking:

_So_, how about this instead, and we Keep It Simple? :smile:

```suggestion
	echo "$version: $fullVersion"
```

Dockerfile-windows.template
tianon: Really good catch, yeah -- I debated!  The `@sh` is just `jq` (https://jqlang.github.io/jq/manual/v1.6/#:~:text=%2C%20%5C%5C%20respectively.-,%40sh,-%3A), but I'm not sure if that exact same escaping is "correct" for PowerShell, so I thought maybe we roll the dice there instead?  I don't really mind either way - it's _probably_ correct enough for PowerShell, especially since it likely won't ever have any actually problematic characters in it. :smile:

edmorley: Yeah makes sense - I'll leave as is then :-)

edmorley:
I've made the those changes - I wouldn't normally force push for review changes, but since this repo uses merge commits rather than squash and merge I wanted to avoid the additional commits ending up on `master`. However, the diff from before can still be seen here:
https://github.com/docker-library/python/compare/7eab18fc115a913534bb0bc5c40bf45c26b61b05..0a4fcf7342f030937b1aad2848d07c0c995bf713

Happy to make further changes if you spot anything else :-)

(And also no rush to merge this; we can wait until the other changes are released first etc)

Dockerfile-windows.template
```diff
@@ -42,45 +42,18 @@ RUN $url = ('https://www.python.org/ftp/python/{0}/python-{1}-amd64.exe' -f ($en
...
-		# get-pip.py installs wheel by default, adding in case get-pip defaults change
+		'setuptools=={{ .setuptools.version }}' \
 		wheel \
-{{ ) else ( -}}
-		--no-setuptools \
-		--no-wheel \
-{{ ) end -}}
-	; \
-	Remove-Item get-pip.py -Force; \
+		; \
```
tianon, line 55:
```suggestion
	; \
```

Dockerfile-linux.template
```diff
@@ -259,63 +259,27 @@ RUN set -eux; \
...
+	python3 --version; \
+{{ if .setuptools then ( -}}
+	\
+	pip3 install \
+		--disable-pip-version-check \
+		--no-cache-dir \
+		--no-compile \
+		{{ "setuptools==\( .setuptools.version )" | @sh }} \
+		wheel \
+		; \
```
tianon, line 272:
```suggestion
	; \
```

tianon (commented):
Last thing I see -- my mental model here is that the closing `;` on a long multi-line is similar to the closing `)` or `}` so I outdent it to make it clear this is the end of the extra long command+args. :+1:

Thanks for your work on this! :heart:

Dockerfile-linux.template
```diff
@@ -176,7 +176,7 @@ RUN set -eux; \
 		--with-lto \
 {{ ) end -}}
 		--with-system-expat \
-		--without-ensurepip \
+		--with-ensurepip \
```
edmorley, line 179: Btw `--with-ensurepip` is the default, so we could also omit this configure option (I left it in to be explicit, but equally happy to remove it).

tianon: Yeah, I'm a fan of being explicit about this -- I wish we could somehow pass in `--default-pip` from here (https://docs.python.org/3/library/ensurepip.html#command-line-interface), but I don't think there's a way. :smile:

edmorley: It looks like the only way would be to switch from `make install` to `make altinstall`:
https://github.com/python/cpython/blob/v3.12.5/Makefile.pre.in#L1903-L1922

...though I haven't dug into the full implications.

tianon: Yeah, I think we're good -- making a symlink isn't too bad, especially when we're already making a bunch. :smile:

edmorley: > It looks like the only way would be to switch from `make install` to `make altinstall`:

And in fact I'm getting confused between `--altinstall` vs `--default-pip` vs the defaults (I don't find the usage text for ensurepip very clear), it's `--default-pip` we need, and there is no make target for that anyway.

Dockerfile-windows.template
edmorley: Indentation fixed in:
https://github.com/docker-library/python/compare/0a4fcf7342f030937b1aad2848d07c0c995bf713..1189682a065093efe1d3ba4095d203c7d2dca9e1

Dockerfile-linux.template
edmorley: https://github.com/docker-library/python/compare/0a4fcf7342f030937b1aad2848d07c0c995bf713..1189682a065093efe1d3ba4095d203c7d2dca9e1

edmorley:
I was going to rebase this on `master` after last night's Python version releases (which have caused conflicts with this PR), however, for some reason the Python 3.12.6 update hasn't been pushed yet by the automation? Did that job fail? (I don't know where those jobs run)
https://github.com/docker-library/python/commits/master/

edmorley:
I've rebased on `master` now that the bump for #960 has landed.

tianon (approved):

tianon:
Also, to be really explicit, thank you so much for bringing up the original thread, being so thoughtful and thorough, seeing it all the way through including the extra cleanup of `get-pip` vs `ensurepip` here, and generally helping respond to the folks that were broken by the whole transition -- it's really appreciated. :heart:

yosifkit (approved):

PR #972, opened 2024-10-07
Add python 3.13 GA release

jnoordsij:
See https://blog.python.org/2024/10/python-3130-final-released.html

Also, drop eol 3.8: https://devguide.python.org/versions/.

Modelled after #868


tianon (approved):
Ahh, giving us the full works :heart:

Thank you!!

PR #974, opened 2024-10-08
Remove deadcode in versions.sh

edmorley:
`minimumSetuptoolsVersion` is unused after Python 3.8 support was removed in #972.


tianon (approved):
Nice catch! :grinning:

Issue #977, opened 2024-10-09
PEP 761: Deprecating PGP signatures


sethmlarson:
Hello, I noticed this project uses PGP signatures and wanted to make you all aware of [my proposed PEP](https://peps.python.org/pep-0761/) for deprecating PGP signatures in future Python versions. Please take a look and leave questions in the [discussion thread](https://discuss.python.org/t/pep-761-deprecating-pgp-signatures-for-cpython-artifacts/67180).

tianon:
Oof, not ideal, but something we can/will adapt to if it's adopted.

Thank you so much for the proactive heads up!  It's really, really appreciated. :heart:

tianon:
The more I look at the details, the more concerned I am about this -- `sigstore` clients are not super generally available (doesn't appear to have pre-existing Debian _or_ Alpine packages, for example, and definitely not in bookworm/stable), especially before we've built/installed Python (which we need to verify before we do :grimacing:), but even if it weren't, installing new packages from PyPI during our build is not something we're comfortable with (we _just_ got away from installing `pip` from PyPI :sweat_smile:). :thinking:

To put that another way, this creates a _new_ verification problem for us in that we need to somehow fetch/verify/trust a `sigstore` client as well. :disappointed:

We'll also need to maintain a mapping of release to release manager + OIDC Issuer (https://www.python.org/downloads/metadata/sigstore/), which is similar to what we already maintain for PGP, so not a huge burden but definitely not easier for us. :sweat_smile:
(Already addressed explicitly in https://discuss.python.org/t/pep-761-deprecating-pgp-signatures-for-cpython-artifacts/67180/6 though, which is nice :heart:)

(Some variant of this might belong in the upstream discussion thread, but I'm mostly trying to write down ideation about how _we_ would implement this, so for now I'm keeping it local.  I do not mind if I get quoted in the upstream thread if something here is useful there, though -- I'm "tianon" there too if that happens and the quoter wants to @ me.)

tianon:
We could use `sigstore-go`, as suggested by the upstream documentation, but we'd have to compile it ourselves because (somewhat ironically) that project doesn't publish precompiled binaries (and solving the "verification" problem there is an interesting/fun case of chicken-and-egg).

tianon:
[...]

PR #981, opened 2024-10-16
Add 3.14-rc variants

sspans-sbp:
Python 3.14 is still in development. This release, 3.14.0a1 is the first of seven planned alpha releases.

https://www.python.org/downloads/release/python-3140a1/

### DIFF
--- a/3.14-rc/alpine3.19/Dockerfile
+++ b/3.14-rc/alpine3.19/Dockerfile
@@ -1,0 +1,125 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
+FROM alpine:3.19
+
+# ensure local python is preferred over distribution python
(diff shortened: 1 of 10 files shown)


sspans-sbp:
I'll update the PR with the correct GPG key in a bit...

jnoordsij:
https://peps.python.org/pep-0761/ seems to confirm that Python is indeed moving from PGP to Sigstore. For >= 3.11 this seems to be in place, so a full migration would not be possible yet; so probably this would require distinguishing versions and implementing two different ways to verify artefacts.

yosifkit:
https://github.com/docker-library/python/issues/977#issuecomment-2403109432

So, I think we'll have to just remove the gpg verification for `3.14` and above and solely rely on the sha256 check (https://github.com/docker-library/python/pull/978) until there is a secure/verifiable way (like an `apt` package in Debian) to get a sigstore client to check the sigstore signature.

tianon:
Updated with an appropriate conditional to remove the PGP logic for 3.14+ :eyes:

yosifkit (approved):

PR #983, opened 2024-10-17
Draft implementation of sigstore for alpine images

sspans-sbp:
All supported python releases now have associated sigstore files available.

This enables sigstore verification in the alpine images using cosign.
Unfortunately cosign 2.4+ is required which is only available in edge, package seems to work fine on 3.19/3.20.

Cosign is removed once the build completes.

### DIFF
--- a/3.10/alpine3.19/Dockerfile
+++ b/3.10/alpine3.19/Dockerfile
@@ -21,14 +21,15 @@ RUN set -eux; \
 		tzdata \
 	;
 
-ENV GPG_KEY A035C8C19219BA821ECEA86B64E628F8D684696D
+ENV CERTIFICATE_ISSUER https://accounts.google.com
+ENV CERTIFICATE_IDENTITY pablogsal@python.org
 ENV PYTHON_VERSION 3.10.15
 ENV PYTHON_SHA256 aab0950817735172601879872d937c1e4928a57c409ae02369ec3d91dccebe79
 
 RUN set -eux; \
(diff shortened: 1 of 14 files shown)


tianon:
Thank you for working on this!  Unfortunately, mixing Alpine releases with packages from Edge is going to be something we're not comfortable with -- it often works, but it also often breaks, and the transition between those is usually unexpected and without warning, so we avoid using Edge entirely as a result. :disappointed:

sspans:
Totally understandable and exactly why I marked this as a draft.
And this independently verified your points mentioned in #977.

However a068d8102dd7988a8634a208c3314aafd21623ff is probably worth merging.

tianon:
> However a068d8102dd7988a8634a208c3314aafd21623ff is probably worth merging.

Can you elaborate?  I left that in when I made #978 because I'd already done the work and it _does_ work, so even if unused, it should be harmless (and I'm still not totally convinced extracting the SHA256 from the sigstore bundles is a great solution to simply getting checksums).  It also gives us a fallback if the sigstore bundles happen to start using a different hash type (although I don't think that's actually a very likely scenario right now).

I'd actually love to improve our confidence in my really hacky extraction of the signature from the sigstore bundles by cross-referencing the SBOM explicitly, but I'd prefer even more to have officially published upstream checksums in a form that's _intended_ for consumption (and then I'd remove _both_ means of scraping a checksum from other data sources).

PR #990, opened 2024-12-05
Add Alpine 3.21 variant

jnoordsij:
Modelled after #925.

Requires https://github.com/docker-library/official-images/pull/18024.

This PR adds a variant for Alpine 3.21 and drops the 3.19 variant at the same time.

See also https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.21.0.


yosifkit (approved):
Thanks!

PR #997, opened 2025-01-07
replace ENV with ARG for build variables

prauscher:
as `ENV` persists in the created images, this clutters environments of many running containers, making printenv more complicated to read and possible making debugging harder.

This PR marks the following variables as build arguments, removing them from scope of the build containers:
* `GPGKEY`
* `PYTHON_VERSION`
* `PYTHON_SHA256`


yosifkit:
Apologies for not responding earlier. 🙇

We use `ENV` on purpose and have no reason to change. Just three "extra" environment variables is quite minimal so we don't mind them persisting in the images. `ARG` makes a hard-to-parse "history" into an impossible-to-parse "history" (`docker history`), so I would never use it. I am open to embedding `GPG_KEY` and `PYTHON_SHA256` directly in the `RUN` line.

Issue #995, opened 2024-12-20
Add -fno-omit-frame-pointer and -mno-omit-leaf-frame-pointer to default compilation flags


artizirk:
Since Python 3.12 it is possible to profile Python programs using [Linux perf profiler](https://perfwiki.github.io/main/) when setting PYTHONPERFSUPPORT=1 environment variable. Unfortunately currently built Python containers are missing `fno-omit-frame-pointer -mno-omit-leaf-frame-pointer` CFLAGS which result in quite useless perf stack traces. It would be great to have those extra CFLAGS added to the image build. That would allow everyone to do low overhead performance profiling on production systems.

* https://docs.python.org/3.12/howto/perf_profiling.html
* https://ubuntu.com/blog/ubuntu-performance-engineering-with-frame-pointers-by-default
* https://fedoraproject.org/wiki/Changes/fno-omit-frame-pointer
* https://www.brendangregg.com/blog/2024-03-17/the-return-of-the-frame-pointers.html

tianon:
Sorry, I'm not sure how I missed this 😭

~~Are there downsides to changing these by default?  Presumably these are omitted by default in the compiler intentionally, and I'd like to make sure we understand why before we disable them (that upstream Python page is great, but the way I read it is "if you want `perf`, you should build this way" not necessarily "it's generally a good idea to build all Python binaries this way" which is a slightly different thing).~~

After typing that, I read your other links (thank you!!) and both Ubuntu and Fedora doing this by default for the whole distro is a pretty strong signal for me. 😄

PR #1000, opened 2025-01-15
Compile with `no-omit-frame-pointer`

artizirk:
This adds support for profiling Python 3.12 and newer using Linux `perf` command.

* https://docs.python.org/3.12/howto/perf_profiling.html

Fixes: #995

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -198,6 +198,15 @@ RUN set -eux; \
 {{ if is_slim or is_alpine then ( -}}
 	LDFLAGS="${LDFLAGS:--Wl},--strip-all"; \
 {{ ) else "" end -}}
+{{
+	# Enabling frame-pointers only makes sense for Python 3.12 and newer as those have perf profiler support
+	if rcVersion | IN("3.9", "3.10", "3.11") then "" else (
+-}}
+# https://docs.python.org/3.12/howto/perf_profiling.html
+# But don't enable frame-pointers on 32bit x86 due to performance drop.
+# https://github.com/docker-library/python/pull/1000#issuecomment-2597021615
+	test "$gnuArch" != 'i686-linux-gnu' && EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"; \
+{{ ) end -}}
 	make -j "$nproc" \
 		"EXTRA_CFLAGS=${EXTRA_CFLAGS:-}" \
 		"LDFLAGS=${LDFLAGS:-}" \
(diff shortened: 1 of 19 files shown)


Dockerfile-linux.template
```diff
@@ -198,6 +198,13 @@ RUN set -eux; \
 {{ if is_slim or is_alpine then ( -}}
 	LDFLAGS="${LDFLAGS:--Wl},--strip-all"; \
 {{ ) else "" end -}}
+{{
+	# Enabling frame-pointer only makes sense for Python 3.12 and newer as those have perf profiler support
+	# https://docs.python.org/3.12/howto/perf_profiling.html
+	if ["3.9", "3.10", "3.11"] | any(.  == rcVersion) then "" else (
+-}}
+	EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"; \
+{{ ) end -}}
```
tianon, line 207: A couple minor suggestions - let's move the doc link into the generated `Dockerfile` so it's easier to find, and some minor `jq` adjustment for a slightly slimmer version test:

```suggestion
{{
	# Enabling frame-pointer only makes sense for Python 3.12 and newer as those have perf profiler support
	if rcVersion | IN("3.9", "3.10", "3.11") then "" else (
-}}
# https://docs.python.org/3.12/howto/perf_profiling.html
	EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"; \
{{ ) end -}}
```

yosifkit:
Ah, these flags are automatic in debug builds based on [supported architectures](https://github.com/python/cpython/blob/3.12/configure#L13046-L13073). Do we also need to limit them to those specific architectures? (`amd64` and `arm64v8`) Or perhaps we follow Ubuntu and only enable it for 64bit platforms? Or maybe some/all arm32vX are fine since they have more spare general-purpose registers than x86/i386?

---

Debian bug for adding these flags to libc: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=767756.

Dockerfile-linux.template
artizirk: Thank you for the feedback. This looks much nicer.

artizirk:
@yosifkit looks like `i386` is the only one that will heavily benefit from **not** using `-no-omit-frame-pointer` compile flag. All other architectures have enough CPU registers for this feature.  

I have added a check for that in the commit.

With i386 support slowly loosing support in Debian[1][2] and 32bit x86 not even being in the CPython supported platforms list[3], it would be interesting to know how may users are still pulling latest i386 Python docker images.

[1]: https://lists.debian.org/debian-devel-announce/2023/12/msg00003.html
[2]: https://lists.debian.org/debian-devel-announce/2024/12/msg00002.html
[3]: https://peps.python.org/pep-0011/

tianon (approved):

PR #1003, opened 2025-02-04
Fix builds for `arm32vX`, `mips64le`, `ppc64le`, `riscv64`, and `s390x`

yosifkit:
By removing the unsupported `-mno-omit-leaf-frame-pointer` flag when building on them

Verifying the supported arches (`amd64` and `arm64v8`):
- https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/x86-Options.html#index-momit-leaf-frame-pointer-2
- https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/AArch64-Options.html#index-momit-leaf-frame-pointer

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -202,10 +202,36 @@ RUN set -eux; \
 	# Enabling frame-pointers only makes sense for Python 3.12 and newer as those have perf profiler support
 	if rcVersion | IN("3.9", "3.10", "3.11") then "" else (
 -}}
+{{ if is_alpine then ( -}}
+		arch="$(apk --print-arch)"; \
+{{ ) else ( -}}
+		arch="$(dpkg --print-architecture)"; arch="${arch##*-}"; \
+{{ ) end -}}
 # https://docs.python.org/3.12/howto/perf_profiling.html
-# But don't enable frame-pointers on 32bit x86 due to performance drop.
 # https://github.com/docker-library/python/pull/1000#issuecomment-2597021615
-	test "$gnuArch" != 'i686-linux-gnu' && EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"; \
+		case "$arch" in \
+{{ if is_alpine then ( -}}
+			x86_64|aarch64) \
+{{ ) else ( -}}
+			amd64|arm64) \
+{{ ) end -}}
+				# only add "-mno-omit-leaf" on arches that support it
+				# https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/x86-Options.html#index-momit-leaf-frame-pointer-2
+				# https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/AArch64-Options.html#index-momit-leaf-frame-pointer
+				EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"; \
+				;; \
+{{ if is_alpine then ( -}}
+			x86) \
+{{ ) else ( -}}
+			i386) \
+{{ ) end -}}
+				# don't enable frame-pointers on 32bit x86 due to performance drop.
+				;; \
+			*) \
+				# other arches don't support "-mno-omit-leaf"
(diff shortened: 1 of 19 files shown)


Dockerfile-linux.template
```diff
@@ -202,10 +202,41 @@ RUN set -eux; \
...
 # https://docs.python.org/3.12/howto/perf_profiling.html
-# But don't enable frame-pointers on 32bit x86 due to performance drop.
 # https://github.com/docker-library/python/pull/1000#issuecomment-2597021615
-	test "$gnuArch" != 'i686-linux-gnu' && EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"; \
+		case "$arch" in \
+{{ if is_alpine then ( -}}
+			x86_64|aarch64) \
+{{ ) else ( -}}
+			amd64|arm64) \
+{{ ) end -}}
```
LaurentGoderre, line 217: Im wondering if these should be in a jq function to make it more readable but other than that, all good

yosifkit: Yeah, I was tempted to do something like [golang](https://github.com/docker-library/golang/blob/985b3b1a072020748a4846fcad475786cfd6e305/Dockerfile-linux.template#L18-L52), but couldn't decide on a good way to embed it.

I was thinking something like the following, but it'll either interrupt the mental parsing of the `RUN` if placed where it is used, or the flags will be far from their context if placed before the `RUN`.
```jq
{{
	def os_arches:
    		if is_alpine then
			{
				x86_64: " -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer",
...
```

Dockerfile-linux.template
```diff
@@ -202,10 +202,41 @@ RUN set -eux; \
...
+				EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer"; \
+				;; \
+{{ if is_alpine then ( -}}
+			x86) \
+{{ ) else ( -}}
+			i386) \
+{{ ) end -}}
+				# don't enable frame-pointers on 32bit x86 due to performance drop.
+				;; \
+			# no default
```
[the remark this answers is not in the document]
tianon: ie:

```suggestion
{{ if is_alpine then ( -}}
			x86) \
{{ ) else ( -}}
			i386) \
{{ ) end -}}
				# don't enable frame-pointers on 32bit x86 due to performance drop.
				;; \
			*) \
				# other arches don't support "-mno-omit-leaf"
				EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer"; \
				;; \
```

Dockerfile-linux.template
```diff
@@ -202,10 +202,36 @@ RUN set -eux; \
...
+{{ if is_alpine then ( -}}
+			x86) \
+{{ ) else ( -}}
+			i386) \
+{{ ) end -}}
+				# don't enable frame-pointers on 32bit x86 due to performance drop.
+				;; \
+			*) \
+				# other arches don't support "-mno-omit-leaf"
+				EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer"; \
```
tianon, line 232: Looking at this one last time before merging, _if_ we come back to this again, we should move this outside the `case` with a simple `if [ "$arch" != i386 ]; then ...` so that the `amd64 | arm64` case can be _just_ the extra `-mno-...` flag.

Dockerfile-linux.template
```diff
@@ -202,10 +202,36 @@ RUN set -eux; \
...
+{{ ) end -}}
 # https://docs.python.org/3.12/howto/perf_profiling.html
-# But don't enable frame-pointers on 32bit x86 due to performance drop.
 # https://github.com/docker-library/python/pull/1000#issuecomment-2597021615
-	test "$gnuArch" != 'i686-linux-gnu' && EXTRA_CFLAGS="${EXTRA_CFLAGS:-} -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"; \
+		case "$arch" in \
+{{ if is_alpine then ( -}}
+			x86_64|aarch64) \
+{{ ) else ( -}}
+			amd64|arm64) \
```
tianon, line 216: Next time we touch this, we should add more whitespace here:

```suggestion
			x86_64 | aarch64) \
{{ ) else ( -}}
			amd64 | arm64) \
```

tianon (approved):

PR #1018, opened 2025-04-08
Bump python versions

yosifkit:
~3.9.22,~ 3.10.17, 3.11.12, 3.12.10, 3.13.3, and 3.14.0a7
Racing the `update.sh` automation because it is slow.

PR #1019, opened 2025-04-09
Update to 3.9.22

tianon:
- https://github.com/python/cpython/issues/132287 is now resolved

(racing the automation again)

- https://github.com/docker-library/python/pull/1018


tianon:
Going to merge now since we'll re-build-test downstream anyhow.

Issue #1011, opened 2025-03-16
Add Docker Images for Python 3.11, 3.12, 3.13, 3.14 on RISCV64 Architecture


nakata5321:
Hello!  

I noticed that **RISCV64** has Docker images available only for older **Alpine Linux** versions of [Python](https://hub.docker.com/r/riscv64/python). However, newer versions **3.11, 3.12, 3.13, 3.14** are currently missing.  

I would like to understand:  
1. **Why are newer versions not supported?**  
2. **Are there any plans to add support?**  
3. **What challenges are preventing implementation?**  

I have a computer running **RISCV64**, and I am willing to help with testing, building, or resolving potential issues.  

I’d be happy to discuss how we can accelerate support for newer Python versions on this architecture.  

Thank you!

nakata5321:
I found this PR - https://github.com/docker-library/python/pull/931
I suppose tests wasn't successful? because now i see that you exclude RISCV architecture [here](https://github.com/docker-library/python/blob/0bd97693f7105bfb6f0ef81020c6d81ade21bc0b/generate-stackbrew-library.sh#L144)

tianon:
If you click @yosifkit's first link, you'll see "take more than 3 hours to build" -- unfortunately that's outside the realm of reasonable for us, so either we need to improve the speed of upstream's build or we need better hardware (and the long tail on each of those is unfortunately quite long).

tuler:
In case anyone needs it, we maintain unofficial `linux/riscv64` builds based on ubuntu 22.04 and 24.04.
https://hub.docker.com/r/cartesi/python/tags

mmoll:
Would https://riscv.builders/ be of service here?

PR #1038, opened 2025-05-05
disable LTO in RISC-V to speed up build

mmoll:
Closes https://github.com/docker-library/python/issues/1011

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -180,9 +180,10 @@ RUN set -eux; \
 		--enable-shared \
 {{
 	# <3.10 does not have -fno-semantic-interposition enabled and --with-lto does nothing for performance
+	# skip LTO on alpine on riscv64: https://github.com/docker-library/python/pull/935, https://github.com/docker-library/python/pull/1038
 	if rcVersion == "3.9" then "" else (
 -}}
-		--with-lto \
+		$(test "$gnuArch" != 'riscv64-linux-musl' && echo '--with-lto') \
 {{ ) end -}}
 		--with-ensurepip \
 	; \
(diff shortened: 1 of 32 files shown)


yosifkit:
After doing some emulated, local tests, I think this seems reasonable to try again. It'll be a less optimized python, but that's better than no riscv64 `python` image. (though users could just install the alpine-provided python3 package in an image `FROM alpine`, but then they are limited to a specific version)

None of these builds are as slow as I was getting in https://github.com/docker-library/python/pull/931 and it's on the same machine. 😕 🤷 
```console
$ # --with-lto
$ for ver in 3.10 3.11 3.12; do time docker build --no-cache --platform linux/riscv64 ./$ver/alpine3.20/; done
real    19m13.374s

real    30m32.618s

real    33m12.560s

$ # without lto
$ for ver in 3.10 3.11 3.12; do time docker build --no-cache --platform linux/riscv64 ./$ver/alpine3.20/; done
real    8m49.526s

real    9m41.161s

real    9m22.683s
```

---

Requested change: also remove [this part](https://github.com/docker-library/python/blob/c2e4187abb7b90c73a287fc748abe5bb53c59a45/generate-stackbrew-library.sh#L139-L144) of `generate-stackbrew-library.sh` to stop removing the riscv64 from the architecture lists.

Dockerfile-linux.template
```diff
@@ -182,7 +182,7 @@ RUN set -eux; \
 	# <3.10 does not have -fno-semantic-interposition enabled and --with-lto does nothing for performance
```
tianon, line 182:
```suggestion
	# <3.10 does not have -fno-semantic-interposition enabled and --with-lto does nothing for performance
	# skip LTO on alpine on riscv64: https://github.com/docker-library/python/pull/935, https://github.com/docker-library/python/pull/1038
```

mmoll:
thanks for the comments, should all be done :bowtie:

tianon (approved):

Issue #947, opened 2024-08-10
Is there a nogil container?


impredicative:
I want a nogil container, also with all optimizations, etc. I assume that the current containers use the GIL.

tianon:
We don't have a "nogil" image currently.  For 3.13 we're using the default compilation settings, which AFAIK still results in the GIL.  Unfortunately, compiling Python is pretty heavy, so expanding our current support matrix to include new "nogil" variants is also going to be a tough sell. :see_no_evil:

tianon:
I'm not sure how you got there from what I said -- perhaps I can clarify: we _do_ compile Python (currently ~42 times _per architecture_, in fact, and we support ~10 architectures where compilation is necessary), and it's _very_ heavy to do so, and adding _more_ variants would require us to do so even _more_ times, so is not something we're willing to consider unless/until it is an officially recommended configuration (ideally the default configuration, but we'd be willing to consider an official statement along the lines of "everyone should try this out").

tianon:
I agree that nogil is the future of Python, and there's nothing strange about the request.  If I were completely opposed to the idea in general, I probably would've closed the issue already.  It's more a matter of resourcing/prioritization, not desire/understanding.  If the demand were actively higher, it would _also_ be a lot easier to justify the added resources/maintenance spend -- the best way for folks to signal that is to react with :+1: on the top post in this thread.

To put that another way, my answer is more accurately "not yet", not "never".

If you'd like to _help_ with the work, figuring out what changes are necessary to our current `Dockerfile`s would be a really useful first step.

ldeluigi:
An experimental JIT-enabled build would also be great to have

sekrause:
Instead of prodiving a whole second set of "nogil" Docker images the free-threaded executable could also be part of the normal images just like it is done with the official Windows and macOS installers:

> CPython now has experimental support for running in a free-threaded mode, with the global interpreter lock (GIL) disabled. This is an experimental feature and therefore is not enabled by default. The free-threaded mode requires a different executable, usually called `python3.13t` or `python3.13t.exe`. Pre-built binaries marked as free-threaded can be installed as part of the official Windows and macOS installers, or CPython can be built from source with the --disable-gil option.

https://docs.python.org/3.13/whatsnew/3.13.html#free-threaded-cpython

So we would have `/usr/local/bin/python3.13t` in every image and call this to use the free-threaded mode.

Unfortunately that still means that all of Python has to be compiled twice.

sekrause:
[...]

PR #1042, opened 2025-05-19
Add Debian Trixie (and remove Bullseye)

tianon:
Closes https://github.com/docker-library/python/issues/1041

Issue #111, opened 2016-05-28
find_library broken in alpine image


rwrnet:
According to http://bugs.alpinelinux.org/issues/5264 its known, that find_library is broken with python3 on Alpine.

However, with python2 it works on pure alpine:3.3, 

```
$ docker run --rm -ti alpine:3.3 sh -c "apk add --no-cache file python && python -c 'from ctypes.util import find_library;print find_library(\"c\")'"
...
libc.musl-x86_64.so.1
```

but not on python:2-alpine,

```
$ docker run --rm -ti --entrypoint sh python:2-alpine -c "python -c 'from ctypes.util import find_library;print(find_library(\"c\"))'"
None
```

so python:2-alpine seems to break something in addition to the general alpine issue, which renders even version 2 broken. Maybe the find_library issue could also be worked around by python image in general.

ushuz:
`find_library` relies on `ldconfig` `gcc` or `objdump` according to its document. `gcc` `objdump` need to be installed (by `apk add gcc`) separately as they are not included in the docker image, but won't work for some libraries, like `libcairo`. I've encountered this myself.

`ldconfig` seems broken according to the author of [this bug](https://bugs.alpinelinux.org/issues/4512), but the maintainer didn't say whether `ldconfig` is broken, instead he provided [a patch](http://git.alpinelinux.org/cgit/aports/tree/main/python/musl-find_library.patch) addressing the issue.

Alpine image may need to apply the same patch to make `find_library` functional. 

@yosifkit @tianon @ncopa

ncopa:
@ushuz `ldconfig` works just fine. it does not cache anything like gnu libc does so there is nothing to print with `ldconfig -p`. as you understand they (rightfully) don't trust the ldconfig output parsing so the fall back to using `gcc` or `objdump` because the likely thought that there is no way there exist any system without those tools installed.

The patch we use in alpine is incomplete. The fundamental problem is when you try to dlopen libc itself, then bad things will happen. upstream musl libc has said they will fix but I don't know what the status is there.

We should probably report this to upstream python.

ushuz:
@ncopa But why calling `ldconfig -p` on `alpine:latest` outputs `Illegal option -p`?

ushuz:
@tianon 

> I'm a little confused here -- what's the use case for find_library('c') in the first place? It seems strange to me to be trying to dlopen libc itself in this way -- aren't most of the functions it provides already available through the Python standard library in some other way?

[...]

PR #1043, opened 2025-05-22
DNM: enable the experimental JIT in 3.13+

tianon:
This is for testing, not for merging.

See:
- https://github.com/docker-library/python/issues/947


tianon:
Oof, there's a lot to unpack with this one.

- https://docs.python.org/3.13/whatsnew/3.13.html#an-experimental-just-in-time-jit-compiler mentions `--experimental-jit-interpreter` but it's actually `--enable-experimental-jit` as noted further down
- https://github.com/python/cpython/blob/3.13/Tools/jit/README.md has several other requirements, notably Python 3.11+ and LLVM version 18 (specifically and only 18: https://github.com/python/cpython/blob/3.13/Tools/jit/_llvm.py#L11)

That is going to be really limiting in our ability to enable this (for example, LLVM 18 isn't even available in Debian stable - it's currently in testing, but I don't know if it'll survive to release of trixie).

tianon:
(and to head it off: compiling LLVM from source is very much out-of-the-question unless/until upstream decides this feature is part of a standard recommended install and should be enabled for ~everyone)

tianon:
Rebased on #1042, applied more hacks; at least the `slim-trixie` builds here should succeed.

tianon:
([unsurprisingly](https://github.com/docker-library/python/issues/111), it looks like Alpine is unsupported: `build.py: error: argument target: invalid get_target value: 'x86_64-pc-linux-musl'`)

tianon:
In the spirit of https://bsky.app/profile/savannah.dev/post/3lskscidafk2e slash https://docs.python.org/3.14/whatsnew/3.14.html#binary-releases-for-the-experimental-just-in-time-compiler, I have updated this to use `yes-off`, which is clearly considered semi-officially OK for released binaries (as the Windows and macOS official releases are now built with it).  Hopefully that's true of 3.13 as well. :see_no_evil:

We do still need to resolve the LLVM version issue ... somehow.
(I really wish https://apt.llvm.org/ included releases, not just nightlies :disappointed:)

PR #1050, opened 2025-06-03
Update to 3.13.4, 3.12.11, 3.11.13, 3.10.18, 3.9.23

tianon:
https://pythoninsider.blogspot.com/2025/06/python-3134-31211-31113-31018-and-3923.html

(racing `update.sh` because these are technically security bumps)


yosifkit (approved):

tianon:
Oop, hit the end of the day so the bot won anyways. :smile:

Issue #1012, opened 2025-03-19
could you please upgrade setuptools version to latest


RohitThakur92:
I am using **python:3.11-alpine3.21** docker image from docker hub however this docker image is having one high level of vulnerability which is related to setuptools library. I would recommend you to upgrade the setuptools version from 65.5.1 to 76.1.0, which can help us to resolve the currently present vulnerability and republish the docker image to dockerhub

edmorley:
These images by design use the pip/setuptools versions that are bundled with the version of Python being installed.

Python 3.11 comes with setuptools v65.5.0:
https://github.com/python/cpython/tree/3.11/Lib/ensurepip/_bundled

So this is expected, and not something that will be changed in these images. (Updating to setuptools 70+ would be a breaking change for a start.)

You will either need to either:
1. Update to Python 3.12 or newer (which no longer bundles setuptools)
2. Update setuptools in your own `Dockerfile`
3. Ask upstream CPython to update to newer setuptools in Python 3.11
4. Suppress the vulnerability alert if appropriate (you didn't say what vulnerability you were referring to, but it's quite possible it's a non-issue in practice - many setuptools codepaths are not used when it's used as a pip build backend, or need several other criteria to be a problem etc)

See also:
- #942
- #952

yosifkit:
> These images by design use the pip/setuptools versions that are bundled with the version of Python being installed.

@edmorley is correct. We do not update `setuptools` in the image; we just include the bundled version. See also https://github.com/docker-library/python/issues/781#issuecomment-1372524712

PR #1051, opened 2025-06-10
upgrade versions of setup tools

sadewa25:
This upgrade will solved the issues of CVE-2025-47273 which had the high score on vulnerability database.


edmorley:
@ns-svemu Ah good spot! Yeah once those upstream changes are released these images should pick that up automatically (and we can then also optionally simplify the version handling in this repo too). I've written some more here:
https://github.com/docker-library/python/issues/1012#issuecomment-2976172015

PR #1094, opened 2025-10-31
Remove all traces of the now-EOL 3.9 release series

tianon:
See https://devguide.python.org/versions/ and https://www.python.org/downloads/ especially:

> end of life, last release was 3.9.25

(Follow-up to https://github.com/docker-library/official-images/pull/20201)


tianon:
On the major plus side, the only affected DOI downstream is `hylang` and it'll automatically adjust :muscle:

PR #1099, opened 2025-12-03
Add Alpine 3.23 and drop 3.21

jnoordsij:
Modelled after #1048.

Requires https://github.com/docker-library/official-images/pull/20400.

This PR updates Alpine to the latest stable version: 3.23.

See also https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.23.0.


tianon (approved):
:heart_decoration:

PR #1107, opened 2026-01-20
Use more defensive `LDFLAGS` append method

### LINKED ISSUE #1106: Wrong parameters from dpkg-buildflags for LDFLAGS on Forky
I'm building CPython on my own CI infrastructure for Forky (Debian 14, testing). Currently the output of `dpkg-buildflags --get LDLAGS` is broken, because:
* it contains a space and breaks further concatenations in the Dockerfile e.g. with `,--strip-all`
* it emits a compiler flag appended to linker flags.

The output I see is:
`LDFLAGS="-Wl,-z-relro -fcf-protection,--strip-all"`

When debugging the output from `dpkg-buildflags`, I get
`LDFLAGS="-Wl,-z-relro -fcf-protection"`

Looking at the outputs of make, I noticed `-fcf-protection` close to other compiler flags, which is correct. Thus, `-fcf-protection` seems to be applied two times.

Workaround:
`LDFLAGS="${LDFLAGS%% *}"`

I'm not sure who is responsible for `dpkg-buildflags` and where to report it.
I just wanted to inform `docker-library` of a potential future problem for upcoming Forky - if not fixed until then.

--------------
I checked Forky and Trixie images from [Docker Hub](https://hub.docker.com/_/debian):
|                      | `CFLAGS`                       | `LDFLAGS`                      |
| -------------------- | ------------------------------ | ------------------------------ |
| Debian 13, Trixie    | `-g -O2 ..... -fcf-protection` | `-Wl,-z,relro`                 |
| Debian 14, Forky     | `-g -O2 ..... -fcf-protection` | `-Wl,-z,relro -fcf-protection` |

Documentation for `-fcf-protection`: https://gcc.gnu.org/onlinedocs/gcc-15.2.0/gcc/Instrumentation-Options.html#index-fcf-protection

This option is not a linker option.

-----
/cc @skoehler

tianon:
In Debian Trixie and older, `dpkg-buildflags --get LDFLAGS` returns something like `-Wl,-z-relro` and in Forky and newer it looks like `-Wl,-z-relro -fcf-protection`, so instead of relying on that `-Wl` and appending to the existing value, we can (and should) just add new `-Wl` values to the end, which _should_ be combined together correctly by the compiler/linker.

- Fixes #1106

### DIFF
--- a/3.10/alpine3.22/Dockerfile
+++ b/3.10/alpine3.22/Dockerfile
@@ -82,7 +82,7 @@ RUN set -eux; \
 # set thread stack size to 1MB so we don't segfault before we hit sys.getrecursionlimit()
 # https://github.com/alpinelinux/aports/commit/2026e1259422d4e0cf92391ca2d3844356c649d0
 	EXTRA_CFLAGS="-DTHREAD_STACK_SIZE=0x100000"; \
-	LDFLAGS="${LDFLAGS:--Wl},--strip-all"; \
+	LDFLAGS="${LDFLAGS:-} -Wl,--strip-all"; \
 	make -j "$nproc" \
 		"EXTRA_CFLAGS=${EXTRA_CFLAGS:-}" \
 		"LDFLAGS=${LDFLAGS:-}" \
@@ -92,7 +92,7 @@ RUN set -eux; \
 	rm python; \
 	make -j "$nproc" \
 		"EXTRA_CFLAGS=${EXTRA_CFLAGS:-}" \
-		"LDFLAGS=${LDFLAGS:--Wl},-rpath='\$\$ORIGIN/../lib'" \
+		"LDFLAGS=${LDFLAGS:-} -Wl,-rpath='\$\$ORIGIN/../lib'" \
 		python \
 	; \
 	make install; \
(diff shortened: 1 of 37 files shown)


tianon:
Verified locally by building with `docker buildx build --pull 3.14/slim-trixie --pull --build-context debian:trixie-slim=docker-image://debian:forky-slim` (and it was successful).

Additionally, a basic test that `--strip-all` is indeed taking effect:

| tag | master | this PR |
| - | - | - |
| `slim-trixie` | [`119MB`](https://github.com/docker-library/python/actions/runs/21103142213/job/60922348719#step:8:21) | [`119MB`](https://github.com/docker-library/python/actions/runs/21184455944/job/60935164839?pr=1107#step:8:21) |
| `slim-bookworm` | [`122MB`](https://github.com/docker-library/python/actions/runs/21103142213/job/60922348792#step:8:17) | [`122MB`](https://github.com/docker-library/python/actions/runs/21184455944/job/60935164939?pr=1107#step:8:17) |

yosifkit (approved):

tianon:
Hah, I created a conflict by merging #1108 first :joy:

PR #1122, opened 2026-06-06
Add builds for Debain forky, a.k.a. testing

lpsinger:


yosifkit:
While we will add Debian Forky-based images eventually, it will not be until its release (https://www.debian.org/releases/). There isn't a release date yet, but it'll likely be sometime in June-August 2027.

tianon:
For users who want to test Forky sooner, doing something like `--build-context debian:trixie-slim=docker-image://debian:forky-slim` (or `--build-context buildpack-deps:trixie=docker-image://buildpack-deps:forky`, as appropriate) while building the images should mostly just work.

Issue #1112, opened 2026-02-20
Rebuild Request: `python:3.10-alpine` — CVE-2026-23949 & CVE-2026-24049 (HIGH/CRITICAL)


tpickle-py:
### Summary

The current `python:3.10-alpine` image ships with outdated versions of `jaraco.context` and `wheel` that contain known, **fixed** vulnerabilities. A rebuild is needed to pick up the patched versions.

Both CVEs have fixed versions available and this should be resolved by a simple image rebuild.

---

### Affected Image

`python:3.10-alpine`

---

### Vulnerabilities

| Library | CVE | Severity | Installed Version | Fixed Version | Description |
|---|---|---|---|---|---|
| `jaraco.context` | CVE-2026-23949 | HIGH | 5.3.0 | 6.1.0 | Path traversal via malicious tar archives |
| `wheel` | CVE-2026-24049 | HIGH | 0.45.1 | 0.46.2 | Privilege escalation or arbitrary code execution via malicious wheel file |

References:
- https://avd.aquasec.com/nvd/cve-2026-23949
- https://avd.aquasec.com/nvd/cve-2026-24049

---

### Impact

These vulnerabilities are being flagged as **Critical** by JFrog Xray, blocking our CI/CD pipeline builds. The fixes are already available upstream — this requires only an image rebuild to pull in the updated package versions.

---

yosifkit:
`setuptools` already is newer than `66.0.0`:
```
$ docker run -it --rm python:3.10-alpine sh
Unable to find image 'python:3.10-alpine' locally
3.10-alpine: Pulling from library/python
530a9815c71f: Pull complete
fef6c4afe209: Pull complete
2ab2d1657c36: Pull complete
9f932b89ad4c: Download complete
2491ffd62a71: Download complete
Digest: sha256:b974a5de91b4ac6da8313502cd5bfe65c499e390d32658e1f2deea26fa5afb14
Status: Downloaded newer image for python:3.10-alpine
/ # pip list
Package    Version
---------- -------
pip        23.0.1
setuptools 79.0.1
wheel      0.45.1
```

[...]

PR #1124, opened 2026-06-15
Pinning wheel at version 0.46.3

caheinz2:
wheel <= 0.46.1 currently has a high severity CVE, and the version pinned in this repo is affected. The pin was originally due to wheel 0.46.0 removing `bdist_wheel`, but it was [later re-added in 0.46.2](https://wheel.readthedocs.io/en/stable/news.html), with a related fix in 0.46.3. Would it be possible to pin at 0.46.3 now, since the version fixes the CVE and won't break earlier versions of setuptools?

Related issues:
[CVE-2026-24049 for wheel <=0.46.1](https://github.com/docker-library/python/issues/1112)
[wheel 0.46.0 release had breaking changes](https://github.com/docker-library/python/issues/1023)

### DIFF
--- a/3.10/alpine3.23/Dockerfile
+++ b/3.10/alpine3.23/Dockerfile
@@ -123,8 +123,7 @@ RUN set -eux; \
 		--no-cache-dir \
 		--no-compile \
 		'setuptools==79.0.1' \
-		# https://github.com/docker-library/python/issues/1023
-		'wheel<0.46' \
+		'wheel==0.46.3' \
 	; \
 	pip3 --version
 
--- a/3.10/alpine3.24/Dockerfile
+++ b/3.10/alpine3.24/Dockerfile
@@ -123,8 +123,7 @@ RUN set -eux; \
 		--no-cache-dir \
 		--no-compile \
 		'setuptools==79.0.1' \
-		# https://github.com/docker-library/python/issues/1023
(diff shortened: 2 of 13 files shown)


caheinz2:
Hey thanks for taking a look at this PR.

I realize that bumping setuptools was an approach that was discussed in the past and would diverge from the `ensurepip` matching strategy. 

I mainly wanted to open this PR because wheel re-added the functionality they previously removed, and I wasn't sure if you all were aware. Unfortunately, just bumping wheel doesn't remove the CVE finding, because setuptools also bundles wheel at an earlier version. I'm not seeing a way to remove the CVE finding without bumping setuptools, but that still may not be acceptable in this repo. I'm happy to close this PR or take another approach if needed, this might not be the right place to bump these versions. Let me know either way!

versions.json
```diff
@@ -25,7 +25,7 @@
       }
     },
     "setuptools": {
-      "version": "79.0.1"
+      "version": "81.0.0"
```
tianon: Bumping setuptools is a no-go, but the wheel bump is probably fine?

caheinz2: Yeah fair enough - I'll revert the setuptools bump.

tianon (approved):

yosifkit (approved):

